PR #4, opened 2015-06-23
SpectralLine Class

profxj:
Here is a 3rd prong..

I appreciate I am adding too much code a bit too fast.  Am somewhat
motivated to have enough in place to proceed with my research.  

Anyways, inspect and suggestion changes.   This is a *major*
aspect of the package so we need to come to agreement on how
best to setup this Class. 

As is, I have 3 aspects to it:

1.  data -- rest wavelength, oscilator strength, species, etc.
2.  analy -- info to perform analysis on the line, e.g. spectrum, wavelength integration range
3.  attrib -- attributes of the line, e.g. redshift, RA/DEC, N, EW, etc.

Each is a dict for now.

Give a look and see what you think.


nhmc:
Thanks, I'll try to look at this by tomorrow.  I don't have much time free today.

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        # dwv
+        dwv = wv - np.roll(wv,1)
+        dwv[0] = dwv[1]
+
+
+        # Simple boxcar
+        EW = np.sum( dwv * (1. - fx) ) 
+        varEW = np.sum( dwv**2 * sig**2 )
+        sigEW = np.sqrt(varEW) 
+
```
nhmc: Use `np.hypot` instead of `sum` and `sqrt`.

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+            raise ValueError('Rest wavelength must be a Quantity.')
+        self.wrest = wrest
+
+        # Other
+        self.data = {} # Atomic/Moleculare Data (e.g. f-value, A coefficient, Elow)
+        self.analy = {'spec': None, # Analysis inputs (e.g. spectrum; from .clm file or AbsID)
+            'WVMNX': [0., 0.], # Wavelength interval about the line (observed)
+            'VLIM': [0., 0.]*u.km/u.s, # Velocity limit of line, relative to self.attrib['z']
+            'FLG_ANLY': 1 # Analyze
+            }
```
nhmc: I suggest changing

WVMNX -> wlim
VLIM -> vlim
FLG_ANLY -> do_analysis

These seem more intuitive to me. I don't think there's any need for capitals.

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        self.attrib = {   # Properties (e.g. column, EW, centroid)
+                       'RA': 0.*u.deg, 'DEC': 0.*u.deg,  #  Coords
```
nhmc: DEC -> Dec

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        # Data
+        self.data.update(self.llist[self.wrest])
+
+        #
+        self.analy.update( {
+            'FLG_EYE': 0,
+            'FLG_LIMIT': 0, # No limit
+            'DATFIL': '', 
+            'IONNM': self.data['name']
+            })
```
nhmc: Please lower case all of FLG_EYE, FLG_LIMIT, DATFIL, IONNM.  Also could we use 'datafile' instead of 'datfil'? And just 'name' or 'ionname' instead of 'ionnm'?

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # Additional attributes for Absorption Line
+        self.attrib.update({'N': 0., 'Nsig': 0., 'flgN': 0, # Column
```
nhmc: maybe change flgN -> flagN?

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        nndt[mask] = np.log(1./fx[mask])*cst
+
+        # Sum it
+        ntot = np.sum( nndt*delv )
+        tvar = np.sum( (delv*cst*sig/fx)**2 )
+
+        # Fill
+        self.attrib['N'] = ntot
+        self.attrib['sigN'] = np.sqrt(tvar)
+        logN, sig_logN = xsb.lin_to_log(self.attrib['N'].value, self.attrib['sigN'].value)
```
nhmc: Need to add lin_to_log to line_tools (we don't want xastropy as a requirement, right?)

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        txt = txt + ']'
+        return (txt)
+
+
+## #################################    
+## #################################    
+## TESTING
+## #################################    
+if __name__ == '__main__':
+
```
nhmc: Need to move these tests to the tests/ directory, otherwise py.test won't run them.

nhmc:
Overall this looks fine to me. Can we enable defining a line by the the name + a rough wavelength. i.e. 

```python
abslin = AbsLine('CIV 1548')
```

I don't want to have to remember rest wavelengths to three decimal places :)

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        # Saturated?
+        satp = np.where( (fx <= sig/5.) | (fx < 0.05) )[0]
+        if len(satp) > 0:
+            mask[satp] = False
+            lim = np.where(sig[satp] > 0.)[0]
+            if len(lim) > 0:
+                sub = np.maximum(0.05, sig[satp[lim]]/5.)
+                nndt[satp[lim]] = np.log(1./sub)*cst
+                flg_sat = len(lim)
```
nhmc: `flg_sat` doesn't seem to be saved anywhere?

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # dv
+        delv = velo - np.roll(velo,1)
+        delv[0] = delv[1]
+
+        # Atomic data
+        cst = (10.**14.5761)/(self.data['fval']*self.wrest) / (u.km/u.s) / u.cm * (u.AA/u.cm)
+
+        # Mask
+        mask = (pix == pix) # True = good
```
nhmc: `pix == pix` is neat, but to me it's clearer to use `mask = np.ones(len(pix), dtype=bool)`. But your call.

linetools/spectralline.py
```diff
@@ -0,0 +1,370 @@
...
+        # For convenience + normalize
+        velo = spec.velo[pix]
+        fx, sig = parse_spec(spec, **kwargs)
+
+        # dv
+        delv = velo - np.roll(velo,1)
+        delv[0] = delv[1]
+
+        # Atomic data
+        cst = (10.**14.5761)/(self.data['fval']*self.wrest) / (u.km/u.s) / u.cm * (u.AA/u.cm)
```
nhmc: Can we add a reference for where the 10**14.5761 comes from, or calculate it using astropy constants? See `Nvel_from_tau` in https://github.com/nhmc/Barak/blob/master/barak/absorb/aod.py for an example.

linetools/spectralline.py
profxj: Are you sure about this one?  

If so, please give me the explicit code.  I fear it isn't
obvious to me.

Thanks.

linetools/spectralline.py
profxj: Done.  Although I used wvlim instead of wlim.

linetools/spectralline.py
profxj: Ok.  I tend to find that these keys become all caps when written
to a FITS table and back.  But, I'll work with it.

linetools/spectralline.py
profxj: Done.

linetools/spectralline.py
nhmc: Oops, I was wrong - I thought `np.hypot` takes more than two inputs, but it
doesn't.  So leave it as is.

linetools/spectralline.py
profxj: Agreed.  I haven't implemented the AODM code yet. 
Sorry about that.

linetools/spectralline.py
profxj: Agreed.  Will put the AODM one in there when I get it done.
Have removed the other two.

linetools/spectralline.py
profxj: Instantiate by name of transition added.

Will work on flg_sat...

linetools/spectralline.py
profxj: Will do.

profxj:
Pushed the edits.

Will move on to AODM next.

linetools/spectralline.py
profxj: Done.  Using datafile and name.

PR #5, opened 2015-06-25
Morton2000

profxj:
Added Morton2000 and made a few other edits.
Added tests for reading Morton files.
Changed Verner1994 to be a FITS table.


linetools/lists/parse.py
```diff
@@ -150,6 +151,13 @@ def read_verner94():
     vot = vo_parse(verner94)
     # Table
     tbl_6 = vot.get_table_by_index(5).to_table(use_names_over_ids=True)
+    '''
```
nhmc: Can you remove this instead of commenting out?

nhmc:
All the notebook examples should (eventually) be moved to docs/examples/. Otherwise this looks fine to merge, when you're ready.

linetools/lists/parse.py
profxj: Gone.

profxj:
Done.  Started an doc/examples folder.
Pushed both notebooks in.
Will push all further example files there.

PR #6, opened 2015-06-27
Line lists

profxj:
Dealt with Isotopes in Morton03 list
Enabled HI line list
Added another test or two
Updated LineList example 'doc'
Added "Am" attribute for Mass Number for Deuterium (and future isotopes)
Added CO line list


linetools/lists/parse.py
```diff
@@ -390,6 +422,9 @@ def mktab_morton03(do_this=False, outfil=None, fits=True):
             outfil = lt_path + '/data/lines/morton03_table2.vot'
         m03.write(outfil, format='votable')
     print('mktab_morton03: Wrote {:s}'.format(outfil))
+    #
+    print('mktab_morton03: Now compressing...')
+    subprocess.call(['gzip', '-f', outfil])
```
nhmc: Might be better to use the python gzip library here?

linetools/lists/parse.py
```diff
@@ -405,13 +440,71 @@ def mktab_morton00(do_this=False, outfil=None):
...
     if outfil is None:
         outfil = lt_path + '/data/lines/morton00_table2.fits'
     m00.write(outfil,overwrite=True)
     print('mktab_morton00: Wrote {:s}'.format(outfil))
+    #
+    print('mktab_morton03: Now compressing...')
+    subprocess.call(['gzip', '-f', outfil])
+
+def update_fval(table, verbose=False):
+    '''Update f-values only
```
nhmc: Why is this function needed? Would be useful to give more info here.

linetools/lists/parse.py
```diff
@@ -267,25 +274,38 @@ def parse_morton03(orig=False, tab_fil=None):
                 #xdb.set_trace()
                 # Line index
                 elmi.append(kk)
-            # ION
+
+            # ISOTOPE and ION
             try:
```
nhmc: Can you leave a comment here that the try-accept is necessary to get around the bad byte in Morton00?

nhmc:
I added some comments. Otherwise, it looks fine to merge.

linetools/lists/parse.py
profxj: Done.  Didn't know about the gzip module

linetools/lists/parse.py
profxj: Added some additional documentation.

linetools/lists/parse.py
profxj: Done.

PR #7, opened 2015-06-28
Spectral line

profxj:
REMOVED: Modified spectralline for AODM analysis.
REMOVED: Added a new test to test_absline_anly
Added ismatch function 
Added "closest" for SpectralLine to grab the closesline in the LineList


linetools/spectralline.py
```diff
@@ -120,8 +124,18 @@ def cut_spec(self, normalize=False):
...
+            # Calculate
+            self.analy['spec'].velo = self.analy['spec'].relative_vel(
+                self.wrest*(1+self.attrib['z']))
+            # Cut
+            velo = self.analy['spec'].velo[pix] 
+            # Return
+            return fx, sig, velo
+        else:
+            # Return
+            return fx, sig, wave
```
nhmc: I am not a fan of changing the return values based on a keyword flag.  Can you return a dictionary instead? i.e.:

```python
return fx, sig, dict(vel=velo, wave=wave)
```

or 

```python
return dict(fx=fx, sig=sig, vel=velo, wave=wave)
```

(I like the first option better)

linetools/spectralline.py
```diff
@@ -242,46 +256,31 @@ def fill_data(self,trans, linelist=None):
                        } )
 
     # Perform AODM on the line
-    def aodm(self, **kwargs): 
+    def aodm(self, flg_sat=None):
```
nhmc: Would it be better to pull this AODM code out of the spectral line class into its own function or module? Might be useful to be able to access it without have to create a SpectralLine.

linetools/spectralline.py
```diff
@@ -291,8 +290,7 @@ def aodm(self, **kwargs):
             if len(lim) > 0:
                 sub = np.maximum(0.05, sig[satp[lim]]/5.)
                 nndt[satp[lim]] = np.log(1./sub)*cst
-                flg_sat = len(lim) 
-                raise ValueError('USE FLG_SAT')
+                flg_sat = True
```
nhmc: `flg_sat` doesn't seem to be stored anywhere.

linetools/spectralline.py
profxj: Done

linetools/spectralline.py
profxj: AODM no longer incoporated in this PR

nhmc: Ok great, please merge when you're ready.

PR #12, opened 2015-06-30
fix test

nhmc:
Fixes a test bug


nhmc:
This is a trivial fix, so I'm merging.

PR #14, opened 2015-07-01
Linelist: all_transition_names_from_line()

ntejos:
LineList all_transition_names_from_line() added.
Simple test passed.

### DIFF
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -223,6 +223,58 @@ class LineList(object):
         ldict['name'] = 'unknown'
         return ldict
 
+    def all_transition_names_from_line(self,line):  
+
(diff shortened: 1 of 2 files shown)


nhmc:
Neat, thanks.

Is there a reason why you chose to return strings and not the actual entries from the linelist? I think this would be better to return a sub-table with all the matching rows from the linelist.

It would also be nice to use a shorter name, like `all_trans()`.

ntejos:
There is no strong reason for retrieving names instead of the full table, although it looks less confusing to me this way. I can easily implement a similar routine for retrieving the full table instead anyways and will do so. Will also shorten the name of the function(s).

Thanks for the input!

profxj:
Good addition.

PR #15, opened 2015-07-02
Fix bug in setwave, add warning, add tests

nhmc:
This fixes a bug (I think - please check carefully) in `setwave`, adds a warning if we guess the wavelengths are log, and adds a new test.  It also removes the assumption that `CRVAL1` and `CDELT1` are 1 if they aren't in the header, which seems a bad idea to me.


profxj:
Looks fine to me.

PR #30, opened 2015-07-18
travis

profxj:
Made 2 modifications to help with Travis (I think).
Pushes specutils dependences into def's.


profxj:
Adding in these simple changes.

PR #31, opened 2015-07-20
Travis

profxj:
A number of small changes to get linetools approved by Travis.
Key one was to sort the set files!


profxj:
Passed Travis!

PR #32, opened 2015-07-20
reads SDSS; small edit to XSpectrum1D.plot

profxj:
Enable reading standard SDSS spectra .fit files (e.g. from DR7).
Put histogram as default in plotting


profxj:
Merging..

PR #33, opened 2015-07-22
xspectrum1d stuff

profxj:
Modified setup of xspectrum1d a little (e.g. its own file)
Add two new "from" methods (mainly to enable direct call)


profxj:
Accepting.

PR #35, opened 2015-07-31
Xspectrum1d

profxj:
Added a bit of kludgy code to handle some XShooter files.
Several minor linelist changes.


profxj:
Passed tests.
Accepting.

PR #36, opened 2015-08-01
Add get_cdelt_dcflag function to simplify reading fits spectra

nhmc:
Add a new function to refactor away duplicate code. This is similar to `get_cdelt()` in barak.spec.


nhmc:
Merging.

PR #37, opened 2015-08-04
Gaussian EW added; error estimation still missing though

ntejos:
spectralline: Gaussian EW added; error estimation still missing though


nhmc:
This look fine to me. I think you can estimate the error from the covariance matrix of the fit:
```
x = g.parameters[0]
y = g.parameters[2]
sigEW = EW * np.sqrt(cov[0,0] / x**2 + cov[2,2] / y**2 + 2 * cov[0,2] / (x*y))
```
where `cov = fit_g.fit_info['param_cov']` is the covariance matrix.

profxj:
I think we should go with the covariance for now.  I have occasionally
calculated both that and a simple boxcar error and taken the higher of the two (to be conservative).

linetools/analysis/utils.py
```diff
@@ -43,3 +43,47 @@ def box_ew(spec):
...
+    """
+    # Grab
+    wv,fx,sig = spec
+
+    # dwv
+    dwv = wv - np.roll(wv,1)
+    dwv[0] = dwv[1]
+    
+    # Fit the data using a Gaussian
+    # Initial guesses
```
profxj: We should have these initial values be an optional input to the method.

linetools/analysis/utils.py
```diff
@@ -43,3 +43,47 @@ def box_ew(spec):
...
+    amp_init = np.mean(fx)/2. #half the mean flux
+    stddev_init = 3*np.mean(dwv) #3 pixels
+    mean_init = np.mean(wv) #half wave range
+    # Model initialization
+    g_init = models.GaussianAbsorption1D(amplitude=amp_init.value, mean=mean_init.value, stddev=stddev_init.value)
+    # Fitting initialization
+    fit_g = fitting.LevMarLSQFitter()
+    # Actual fit
+    g = fit_g(g_init, wv, fx, weights=1./sig)
```
profxj: I think we should query whether any sig values are <=0 (or NAN) and
mask those, e.g.

mask = (sig > 0.) & (np.isfinite(sig))
g = fit_g(g_init, wv[mask], fx[mask], weights=1./sig[mask])

linetools/analysis/utils.py
```diff
@@ -43,3 +43,47 @@ def box_ew(spec):
...
+
+    # Area under curve of Gaussian is [amplitude*stdev*sqrt(2*pi)]
+    EW = g.amplitude.value * g.stddev.value * np.sqrt(2 * np.pi) #unitless
+    EW = EW * wv.unit #add the same unit as wv
```
profxj: If the fit (g) has units, why take them out and put them back in?

ntejos: It seems that these implemented Gaussian models from astropy.modeling do not support units...

linetools/analysis/utils.py
ntejos: Will do.

linetools/analysis/utils.py
ntejos: Will do.

nhmc:
Was the last commit added by mistake? No worries if not, if you can just fix the test failure I think it looks ok to merge.

profxj:
Merging.

PR #38, opened 2015-08-06
Line lists

profxj:
Added new EUV lines to line lists.  Will get original data soon.
Also a small edit to the splice method.


nhmc:
This looks good to me. Tests pass, so I'm merging.

PR #39, opened 2015-08-12
Add a continuum to XSpectrum1D, support UVES_popler files, other tweaks

nhmc:
This adds a continuum to XSpectrum1D, and support for UVES_popler files. The XSpectrum1D continuum can be saved (read again) in fits format. There are also a couple of minor changes to XSpectrum1D.plot and reading ascii files.


linetools/spectra/io.py
```diff
@@ -377,9 +401,34 @@ def chk_for_gz(filenm):
...
+
+def parse_UVES_popler(hdulist):
+    """ Read a spectrum from a UVES_popler-style fits file.
+    """
+    from linetools.spectra.xspectrum1d import XSpectrum1D
+
+    hd = hdulist[0].header
+    uwave = setwave(hd) * u.Angstrom
+    co = hdulist[0].data[3]
+    fx = hdulist[0].data[0] * co  #  Flux
```
profxj: Do you really mean to multiply here?
I should have thought it would be divide.

linetools/spectra/xspectrum1d.py
```diff
@@ -111,24 +113,31 @@ def add_noise(self,seed=None,s2n=None):
...
         conti: numpy array
-          Continuum
+          Continuum. Use XSpectrum1D.co if None is given
         verbose: bool, optional (False)
         no_check: bool, optional (False)
           Check size of array?
         """
         # Sanity check
+        if conti is None:
+            if hasattr(self, 'co') and self.co is not None:
```
profxj: Isn't self.co initialized to a bunch of NANs?
Maybe we should initialized to ones.

linetools/spectra/io.py
```diff
@@ -212,30 +237,29 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, flux_tags=None,
...
-        co = np.nan*np.ones(npix)
+    if not hasattr(xspec1d, 'co'):
+        xspec1d.co = co
+        # Final check for continuum in a separate file
+        if co is None and specfil.endswith('.fits'):
+            try:
+                xspec1d.co = fits.getdata(specfil.replace('.fits', '_c.fits'))
+            except IOError:
+                npix = len(xspec1d.flux)
+                xspec1d.co = np.nan * np.ones(npix)
```
profxj: Might be safer to initialized to simply ones.

linetools/spectra/xspectrum1d.py
```diff
@@ -136,7 +145,9 @@ def normalize(self, conti, verbose=False, no_check=False):
                 raise ValueError('normalize: Continuum needs to be same length as flux array')
 
         # Adjust the flux
-        self.flux = self.flux / conti
+        self.flux /= conti
+        if self.uncertainty is not None:
+            self.uncertainty.array /= conti
```
profxj: Should we have a flag that specifies whether the flux has been normalized previously, i.e. to avoid doing this step twice?
Or should we put that on the user.  I guess the latter is fine..

nhmc: I'm inclined to let the user deal with this for now...

linetools/spectra/io.py
nhmc: How about we just leave it as None if no continuum is found?

linetools/spectra/io.py
nhmc: This makes fx equal to the un-normalised flux. I prefer this to be the default - i.e. flux should be un-normalised, and the user can normalise it if desired. But if you strongly prefer the opposite I'm happy to do that too.

profxj: Ok, I understand now.  It is read in normalized.  I was assuming that
Popler wrote it as un-normalized.  This is fine.

linetools/spectra/io.py
profxj: I am fine with None.

profxj:
Make the None edit as above and merge away!

nhmc:
Tests pass, so merging.

PR #40, opened 2015-08-14
Line lists

profxj:
Added gamma values for HI (for Voigt).
Updated names of Deuterium lines in Morton 2003
Access 'gd_lines' by name
And related tests.
Now, a few bits and pieces in XSpectrum1D.  Sorry..


ntejos:
Looks good to me; suggested changes to X seem successfully implemented. Merging.

PR #41, opened 2015-08-18
Add basic interactive plotting

nhmc:
This adds basic classes to allow adding commands for interactive plotting.  XSpectrum1D.plot is tweaked enable XIDL-style keypress navigation, which I much prefer to the standard matplotlib navigation tools.

I plan to extend this with an interactive continuum fitting class soon.

Any comments welcome!


profxj:
This looks real good, Neil.  I should add that I've implemented similar code
in my own xastropy Guis and would need to find a way to meld to the two 
if/when I bring in XSpecGui.  But that is more than doable.

PR #45, opened 2015-08-20
Fix a regression from a previous PR, and add test

nhmc:
Fix a change which was incorrectly added to PR #41 that breaks xastropy. Also add a test.


profxj:
Thanks for fixing.

PR #48, opened 2015-08-25
LineList: reset_data optional key added to subset_lines() by JXP

ntejos:
LineList: reset_data optional key added to subset_lines() by JXP; this is useful for changing subset without reloading data from master tables.


nhmc:
Could you add some more text in the docstring to describe why you would want to do this? (ie. reset _data in the Table).

ntejos:
Requested extra docstring added; waiting for your approval to merge. Otherwise happy to address any other potential concern with this.

nhmc:
Thanks - merging.

PR #50, opened 2015-08-28
spectra/convolve.py: barak convolution routines replaced by astropy.c…

ntejos:
spectra/convolve.py: barak convolution routines replaced by astropy.convolution equivalents.

### DIFF
--- a/linetools/spectra/convolve.py
+++ b/linetools/spectra/convolve.py
@@ -1,7 +1,7 @@
 """ Functions related to convolution.
       Taken from Barak by JXP
       May replace with scipy functions
-      NT: replace barak routines with astropy.convolvolution equivalents
+      NT: replace barak routines with astropy.convolution equivalents
 """ 
 
 # py2.6+ compatibility
--- a/linetools/spectra/convolve.py
+++ b/linetools/spectra/convolve.py
@@ -1,14 +1,16 @@
 """ Functions related to convolution.
       Taken from Barak by JXP
       May replace with scipy functions
+      NT: replace barak routines with astropy.convolvolution equivalents
 """ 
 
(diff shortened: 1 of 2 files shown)


ntejos:
It seems that the new implementation failed test_gauss_smooth() due to a difference of ~1e-3 whereas a precision closer to 1e-7 was required. If Neil and JXP are fine with me updating the test_gauss_smooth() function to pass with the new convolution implementation I will do so.

Also, we should consider to deprecate our concolve.py module and simply use astropy.convolution directly instead.

linetools/spectra/convolve.py
```diff
@@ -1,14 +1,16 @@
...
 """ 
 
 # py2.6+ compatibility
 from __future__ import print_function, absolute_import, division, unicode_literals
 
 import numpy as np
+from astropy.convolution import convolve, Gaussian1DKernel, CustomKernel
 
-def convolve_psf(a, fwhm, edge='invert', replace_nan=False, debug=False):
+def convolve_psf_old(a, fwhm, edge='invert', replace_nan=False, debug=False):
```
nhmc: I think it's better to remove the old versions completely (`convolve_psf_old` and `convolve_window_old`).

linetools/spectra/convolve.py
```diff
@@ -1,14 +1,16 @@
 """ Functions related to convolution.
       Taken from Barak by JXP
       May replace with scipy functions
+      NT: replace barak routines with astropy.convolvolution equivalents
```
nhmc: Typo

nhmc:
Apart from the inline comments, this looks great to me.  Thanks!  Feel to merge when you're ready.

PR #51, opened 2015-08-31
Xspectrum1d

profxj:
Very minor edits.


profxj:
These were simple merges..

PR #54, opened 2015-09-02
Interactive continuum fitting 

nhmc:
This adds the ability to interactively fit a continuum to a spectrum, inspired by the fitcont code from Barak. There is an ipython notebook giving example usage.

I'm not sure how to add tests for interactive code like this - any suggestions are welcome.

### DIFF
--- a/linetools/spectra/xspectrum1d.py
+++ b/linetools/spectra/xspectrum1d.py
@@ -19,6 +19,11 @@ import linetools.spectra.io as lsio
 
 from .plotting import get_flux_plotrange
 
+from ..analysis.interactive_plot import InteractiveCoFit
+from ..analysis.continuum import prepare_knots
+from ..analysis.continuum import find_continuum
+
+
 try:
     from specutils import Spectrum1D
 except ImportError:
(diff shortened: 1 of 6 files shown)


profxj:
Glad to see this added!  There are commercial packages like Travis for GUIs
but are beyond our budget. :)
Looking forward to giving this a shot.

linetools/spectra/xspectrum1d.py
```diff
@@ -526,3 +531,85 @@ def write_to_fits(self, outfil, clobber=True, add_wave=False):
...
+                nchunks = max(3, (wmax - wmin) / float(dw))
+                edges = np.linspace(wmin, wmax, nchunks + 1)
+    
+        if knots is None:
+            knots, indices, masked = prepare_knots(
+                wa, self.flux.value, self.uncertainty.array, edges)
+        else:
+            knots = [list(k) for k in knots]
+    
+        co = (self.co if hasattr(self, 'co') else None)
```
ntejos: This raises an error in the continuum fitting, when self.co = None (i.e. self.co is defined but its value is None). 

Solutions:
- Initialize self.co within self.fit_continuum()
- Initialize self.co at the XSpectrum1D() initialization stage

My preference is for the second option, using something like 

self.co = self.flux * float(X)

so it is the same type and units as self.flux and has the same units. float(X) could be 0. to emphasize that no continuum estimation exists, but it could also be a fully automatic rough estimation.

PR #55, opened 2015-09-02
LSF: module to deal with LSFs of various instruments

ntejos:
Currently it is only implemented for HST/COS
- Deals with arbitrary LSFs
- Deals with the different LSFs for a given HST/COS life time position and central wavelength of the gratings


linetools/spectra/lsf.py
```diff
@@ -0,0 +1,295 @@
...
+        elif len(wv_array.shape) != 1:
+            raise SyntaxError('`wv_array` must be of shape(N,), i.e. 1-dimensional array')
+        
+        #define useful quantities
+        wv_min = np.min(wv_array)
+        wv_max = np.max(wv_array)
+        wv0 = 0.5 * (wv_max + wv_min)
+        
+        lsf_tab = self.interpolate_to_wv0(wv0)
+        f = interp1d(lsf_tab['wv'],lsf_tab['kernel'],kind='cubic',bounds_error=False,fill_value=0)
```
nhmc: You may want to consider Akima splines in linetools.analysis.interp instead of 1-d cubic splines.  It probably doesn't make much difference, though
.

linetools/spectra/lsf.py
```diff
@@ -0,0 +1,295 @@
...
+
+    """
+
+    def __init__(self, instr_config):
+        #mandatory keys for characterizing a spectrograph mode
+        self.mandatory_dict_keys = ['name']
+                
+        #Check correct format
+        if not isinstance(instr_config,dict):
+            raise SyntaxError('`instr_config` must be a dictionary.')
```
nhmc: I don't think `SyntaxError` is the right thing to raise here. `TypeError` would be more appropriate (https://docs.python.org/2/library/exceptions.html#exceptions.TypeError). Or just `RuntimeError` if you don't want to have to worry about it.

nhmc:
This looks excellent to me. I'm happy for you to merge when you're ready.

linetools/spectra/lsf.py
ntejos: Akima looks good and robust to me for the LSFs; so I will do so. Just in case will allow two kinds of interpolation: ('cubic' and 'Akima'). I guess Akima interpolation should also be preferred for QSO continuum fitting (if not already implemented).

PR #57, opened 2015-09-03
Metadata io

nhmc:
Enable reading and writing of the metadata dictionary `Spectrum1D.meta` to FITS files.  It's stored as a json-serialized string in the primary FITS header under a keyword called `METADATA`.

This seems a simpler solution than making a new FITS key-value pair for every metadata entry. It also allows saving more complex things than the FITS keyword-value pairs allow, like lists of lists.

My motivation for doing this is that I want to save a series of continuum spline knots, which are a list of x,y pairs, in a spectrum's metadata.


profxj:
Looks reasonable.  Are we sure it meets the FITS standards?
If so, please merge.

nhmc:
Yes, I'm pretty sure it does. METADATA is not a reserved keyword. 

It requires the CONTINUE syntax, but that's a registered FITS convention since 2007 (http://fits.gsfc.nasa.gov/registry/continue_keyword.html), and it's supported by cfitsio, IDL and astropy.io.fits.

PR #58, opened 2015-09-04
Make "file not found" error message in readspec more infomative

nhmc:


nhmc:
This is a simple change, so merging.

PR #60, opened 2015-09-04
Allow reading of spectra without an error array

nhmc:
This allows reading of spectra without an error array. It includes a test.


profxj:
Looks good.  Merging..

PR #61, opened 2015-09-04
Interactive continuum fit: added `M` key event to move nearest knot g…

ntejos:
Interactive continuum fit: added `M` key event to move nearest knot guessing the y position.

### DIFF
--- a/linetools/analysis/interactive_plot.py
+++ b/linetools/analysis/interactive_plot.py
@@ -230,6 +230,7 @@ A        : add a new spline knot, and use a flux median to guess y position
 _        : halve the number of spline knots
 d        : delete the nearest knot
 m        : move the nearest knot
+M        : move the nearest knot, and use a flux median to guess y position
 
 q        : quit
 """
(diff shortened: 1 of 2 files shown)


linetools/analysis/interactive_plot.py
```diff
@@ -490,6 +491,23 @@ def on_keypress(self, event):
...
+                return
+            # Move a point, estimating via median
+            contx,conty = zip(*self.ax.transData.transform(self.contpoints))
+            sep = np.hypot(event.x - np.array(contx),
+                           event.y - np.array(conty))
+            ind = np.argmin(sep)
+            if ind in (0, len(sep) - 1):
+                return
+            x = event.xdata
+            if not self.contpoints or x not in zip(*self.contpoints)[0]:
```
nhmc: This should just be `if x not in zip(*self.contpoints)[0]:`

linetools/analysis/interactive_plot.py
```diff
@@ -490,6 +491,23 @@ def on_keypress(self, event):
             self.contpoints[ind] = event.xdata, event.ydata
             self.contpoints.sort()
             self.update()
+        elif event.key == 'M':
+            if not between(event.xdata, self.wmin, self.wmax):
```
nhmc: We may also need to add a check that len(self.contpoints) > 0

(I forgot about this)

nhmc:
Thanks for this. I think it's easiest if I merge this as is, and then I'll incorporate my inline comments into PR #62.

PR #62, opened 2015-09-06
Bug fixes to interactive continuum fitting, and add tests

nhmc:
This fixes the bug pointed out by @ntejos in #54, and some others. It also adds basic tests.


nhmc:
I've removed the tests for now, because they don't seem to work with Travis. (I get an error about the DISPLAY variable not being set).

nhmc:
I am merging this, because it fixes a major bug with the interactive continuum fitting should not affect other code.

PR #63, opened 2015-09-08
LineList: minor tweaks

ntejos:
LineList: minor tweaks; basically strongest_transitions() and available_transitions() can now specify n_max (or n_max_tuple) as None, meaning retrieve everything.


nhmc:
This looks great, thanks.

PR #64, opened 2015-09-14
Xspectrum1d

profxj:
Added a write_ascii method.
Currently, it defaults do a ecsv file which are
nice for capturing header information.  Requires yaml
which I added to Travis (on my 3rd try!).
Also added a test.


nhmc:
Looks good. Merging.

PR #65, opened 2015-09-23
edit to maintain table input to readspec

profxj:
I should probably add a test for this one of these days..

Note: Coverage tests aren't working now due to an upgrade in
the Travis software.  Hopefully the astropy folks will put in a fix
that we can copy soon enough.


nhmc:
Looks good, thanks.

profxj:
Merging.

PR #67, opened 2015-10-01
Add plt.show to XSpectrum1D.plot

nhmc:
This adds a `matplotlib.pyplot.show()` command to XSpectrum1D.plot so that the plot will always appear automatically, even when running from a script. This can be disabled with a new show keyword.


profxj:
Thanks for adding this.

PR #68, opened 2015-10-12
Add a test for reading spectra from an astropy Table, plus some docstring improvements

nhmc:
Add a test to avoid a future regression similar to #65. Also tweaks some docstrings to display well in command line ipython.


profxj:
Merging

PR #69, opened 2015-10-12
Make compatible with Python 3

nhmc:
We need to be Python 3 compatible if we want to be taken seriously by the astropy crowd :)

Therefore, this PR makes the changes necessary for linetools to work with Python 3.  All the tests pass in Python 3.4, and hopefully (!) Travis will now run the test suite using that python version, in addition to Python 2.7.

Note that there may still be python 3-related incompatibilities that aren't shown up by our tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ matrix:
     include:
 
         # Do a coverage test in Python 2.
-        - python: 2.7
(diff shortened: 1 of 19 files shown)


profxj:
Thanks so much for doing this.

PR #71, opened 2015-10-13
Linelist cache

nhmc:
Reading linelists can be slow if you need to do it several times in a single python session. This minimises re-reading lists (like 'Strong', 'ISM', etc.) by storing them in a dictionary.

### DIFF
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -10,7 +10,7 @@ except NameError:
     basestring = str
 
 import numpy as np
-import os, imp
+import os
 import copy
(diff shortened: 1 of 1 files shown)


profxj:
I'm not sure what the 'stanard' is for using Globals, but I like having
this functionality.  Should we make it optional, i.e. one has to turn
on the Caching?  Or, alternatively, something that can be disabled?

Aside from that I am quite happy with this.

nhmc:
I added a keyword that can disable using the cache. By default caching is used.

PR #72, opened 2015-10-15
Line lists

profxj:
Took LineList out of AbsLine object  
  -- Didn't make sense to pack that object in this other
Some other abs_line edits

Took FeII lines from Howk00 out of the standard list given 
  input from Y. Zheng


nhmc:
It looks like test_updfval() in lists/tests/test_init_linelist.py also needs to be updated.

profxj:
Updated the test and a few other minor things.

Should be good to go.

PR #73, opened 2015-10-15
Xspectrum1d

profxj:
Edited documentation

Requiring (explicilty) that wvobs be a Quantity in call to relative_vel.


nhmc:
Thanks.

PR #74, opened 2015-10-29
Abund

profxj:
Solar abundances (Asplund 2009 for now)
elements.py  (2015 version)
Some other ions things

Made changes to LineList.available_transitions

Tests too


linetools/abund/solar.py
```diff
@@ -0,0 +1,127 @@
...
+l_path = imp.find_module('linetools')[1]
+
+#
+class SolarAbund(object):
+    '''Class to handle simple Solar Abundance calculations
+
+    Parameters:
+    ----------
+    ref: str, optional
+       'Asplund2009' :: AA&RA meteoritic table (several photometric)
```
nhmc: Can you add the precise reference here? (Asplund et al. 2009, ARA&A, 47, 481)

nhmc:
Thanks for this.

I see you've used a similar class structure to LineList.  After using that for a while now, I find the `__getattr__`  overloading a bit awkward and non-intuitive. It's also potentially confusing if there's a column with a name that matches one of the class methods. Can we simply store the table in an attribute `data` (instead of `_data`) and then if you want to access the table columns directly, use `SolarAbund.data['colname']`?

Overloading `__getitem__` a bit counterintuitive for me. Whenever I access something as an item, I expect there to be a single set of homogeneous keys with similar types of values.  Having different kinds of keys (tuples vs strings, and here also ratios) goes against this expectation. It would make more sense to me to have a function .get() which could take either a tuple or string, and a different function .get_ratio() to find the ratio.

Some of this might just be down to personal preference. @ntejos - what do you think?

profxj:
I agreed, mainly. 

Got rid of the _get_attr overload.  
Certainly not necessary for this class.

And I made get_ratio() a method.  I do, however,
prefer to keep _get_item and overload it.   This just
seems a more natural way to go to me.

nhmc:
That looks good to me. Feel free to merge when ready.

profxj:
Merging

PR #75, opened 2015-11-02
Voigt profile

profxj:
Added code to generate Voigt profile.  

Main routine is through the SpectralLine class although the
core routines are in analysis/voigt.py

Added tests
Added Notebook

Default is to test whether the wavelength array is sampled
at finer than 1/6 the minimum b-value (in km/s).  This
can be turned off (skip_wveval=True).   Anyhow I open
to other suggestions on this.


linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+from astropy import constants as const
+from astropy.modeling import FittableModel, Parameter
+
+from linetools.spectra.xspectrum1d import XSpectrum1D
+from linetools.spectralline import AbsLine
+
+#from xastropy.xutils import xdebug as xdb
+
+# The standard King model
+def voigtking(vin,a):
```
nhmc: This needs a reference, if possible.
It's from Ryan's Alis code, right?  (Is it a translation of the vpfit code?)

linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+# The primary call
+def voigt_model(iwave, line, fwhm=None, flg_ret=1, debug=False, 
+    skip_wveval=False):
+    '''Generates a Voigt model from a line or list of lines
+    This will run *slowly* for many many lines
+
+    Parameters:
+    ------------
+    wave: Quantity array
+      Observed wavelengths
```
nhmc: It would also be nice to be able to generate the profile on an array of velocities. I'm not sure if this can be easily incorporated here, or if it needs a different function.

linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+      If False, the median dwave value is calculated and compared against
+      1/10 of the b-value of the input line(s).  
+      If necessary, the wavelenght array is rebinned for the calculation
+      and the final array is rebinned to the original.
+    fwhm: float, optional [None]
+      FWHM for Gaussian smoothing (pixels)
+    flg_ret : int (1)  Byte-wise Flag for return
+      1: vmodel [XSpectrum1D]
+      2: tau
+      4: Absorbed flux [np.ndarray]
```
nhmc: Can these flags be made into strings to make them easier to remember?

Maybe change the keyword to just 'ret' and make the options 'vmodel', 'tau' and 'flux'/

linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+                    print(iline)
+                tau += voigt_model(wave, iline, fwhm=0., 
+                    flg_ret=2, debug=debug, skip_wveval=True) 
+        else:
+            raise ValueError('voigt: Unknown type for voigt line in your list')
+    else: 
+        raise ValueError('voigt: Unknown type for voigt line')
+
+    # tau
+    if 'tau' not in locals(): # Deals with multiple line input
```
nhmc: Move this block (150-169) up inside the `if` block starting at line 126, instead of testing here for tau in locals().

linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+    # Convolve
+    if fwhm is not None:
+        vmodel.gauss_smooth(fwhm=fwhm)
+    else:
+        warnings.warn('Assuming infinite spectral resolution, i.e. no smoothing.')
+        warnings.warn('Set fwhm to smooth.')
+    
+    # Return
+    ret_val = []
+    if flg_ret % 2 == 1: ret_val.append(vmodel)
```
nhmc: Like I said above, I would prefer to use strings as flags instead of integers. I know that means we can't combine different flags to return multiple things (i.e. flg_ret=5 returns both the vmodel and flux).

Maybe allow multiple strings to return multiple values? i.e. ('tau', 'flux', 'vmodel')?  Then the order of the returned values would also be clear. Currently you have to work out the return order by looking at the code, or trial and error.

linetools/spectralline.py
```diff
@@ -324,6 +324,40 @@ def fill_data(self,trans, linelist=None, closest=False):
...
+    # Voigt
+    def generate_voigt(self, wave=None, **kwargs):
+        """  Generate a Voigt profile model for the absorption line
+        in a given spectrum.
+
+        Parameters:
+        ----------
+        wave: Quantity array
+          Wavelength array on which to calculate the line
+          Must be set if self.analy['spec'] is not filled
```
nhmc: Would be nice to have a velocity array option.

linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+        v0[nl] = 2*v[nl]*10.0
+        p=np.int_(v0[nl])
+        p1=p+1
+        p2=p+2
+        x=0.5*np.int_(v0[nl])
+        y=x+0.5
+        z=x+1.0
+        v1 = v0[nl] * 0.5
+        voigt_prof[nl] = 2.0*((v1-y)*(v1-z)*(h0[p]+a*(h1[p]+a*(h2[p]+a*h3[p]))) - (v1-x)*(v1-z)*2.0*(h0[p1] + a*(h1[p1]+a*(h2[p1]+a*h3[p1]))) + (v1-x)*(v1-y)*(h0[p2] + a*(h1[p2]+a*(h2[p2]+a*h3[p2]))))
+    del nl, nh
```
nhmc: Remove this line

nhmc:
There is an xdb.xplot at the end of the notebook example which should be removed, or converted to use vanilla matplotlib.

linetools/analysis/voigt.py
```diff
@@ -0,0 +1,203 @@
...
+        if isinstance(line,AbsLine):  # Single line as a AbsLine Class
+            minb = line.attrib['b']
+        else:
+            minb = np.min(np.array(
+                [iline.attrib['b'].value for iline in line]))*u.km/u.s
+        # Calculate dwave
+        dwave = np.median(np.abs(iwave-np.roll(iwave,1)))
+        medwave = np.median(iwave)
+        warnings.warn('Using a sub-grid wavelength array because the input array is too coarse.')
+        warnings.warn('Will return values rebinned to the input array.')
```
nhmc: This warning should be inside the if block below, right?

nhmc:
I'm worried this will be very slow, because it uses units a lot. If so, it might be sped up by pulling out the unit conversions, or even requiring a wavelength array that is not a quantity. I will test against the barak equivalents to compare speed, but not until the weekend.

It would be nice to be able to generate tau on a velocity array.  I'm not sure if this should be done in this function, or a different one.

There are also several options for the actual voigt generate code (currently voigtking), and it would be nice to know we have the fastest/most accurate one.  Other options are using scipy's wofz, which is nice because it's one line of code (see https://github.com/nhmc/Barak/blob/master/barak/absorb/voigt.py#L1021), and my version (https://github.com/nhmc/Barak/blob/master/barak/absorb/voigt.py#L1063, which is based on some c code by Joe Liske). We can worry about this last point after the PR is merged though.

linetools/analysis/voigt.py
profxj: Added.

linetools/analysis/voigt.py
profxj: I'm not sure how you would wish to have this implemented.
As such, this one is all yours! ;)

linetools/analysis/voigt.py
profxj: Modified.

linetools/analysis/voigt.py
profxj: I assume you meant the 'del' line.  Gone.

linetools/analysis/voigt.py
profxj: Right

linetools/analysis/voigt.py
nhmc: Fair enough :). I'm happy to use voigt_tau() with a log-linear wavelength array for now.

nhmc:
The new version looks great, thanks for all the changes!

The example notebook needs to be updated (and the final xplot line in it still needs to be removed/changed). We should also test the voigt models against a high-precision voigt profile.  But these can be done later; feel free to merge when ready.

profxj:
Shoot!  I was editing my Voigt notebook in xastropy.

There was some good stuff in there (check out the speed and
accuracy ‘tests’), pushing now and merging once the tests pass.

Cheers,

X

> On Nov 9, 2015, at 4:37 PM, Neil Crighton <notifications@github.com> wrote:
> 
> The new version looks great, thanks for all the changes!
> 
> The example notebook needs to be updated (and the final xplot line in it still needs to be removed/changed). We should also test the voigt models against a high-precision voigt profile. But these can be done later; feel free to merge when ready.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/linetools/linetools/pull/75#issuecomment-155245782>.
> 
> !DSPAM:826,56413c5315731813823668!

----------------------------------------------
J. Xavier Prochaska
UCO/Lick Observatory
1156 High St.
UC Santa Cruz
Santa Cruz, CA 95064

http://www.ucolick.org/~xavier/
http://imps.ucolick.org/
http://www.qpqsurvey.org/
http://k1dm3.ucolick.org/
http://www.ucolick.org/~xavier/IDL/index.html
https://github.com/xastropy/xastropy
https://github.com/linetools/linetools

831-459-2135 (Direct)
831-459-2991 (UCO/Lick Main)
831-459-5244 (Fax)

nhmc:
Neat, thanks!  So wofz is fast, accurate and concise. That doesn't happen often...  

I guess a c/fortran version of voigtking would be much faster, though.

profxj:
Merging.

PR #77, opened 2015-11-16
Add python 3.5, numpy 1.10 to travis testing

nhmc:
It also adds a workaround for this issue astropy/specutils#123.


profxj:
Thanks for dealing with this.  I've posted a few times to astropy that
this needs to be addressed.  Have been sticking with numpy 1.9 for now.

On that note, did you check things are still ok with 1.9?  If so, please merge.

nhmc:
We still test for python 2.7 and numpy 1.9. Merging.

PR #79, opened 2015-11-17
linet_absline script for plotting a simple absorption line

profxj:
Introduces simple scripts to linetools.  Not sure we want to go there (if not, I'll push this
  into xastropy).

Didn't create a test as I'm not sure what I'd do.  The functionality is a mix
 of things that are tested already.

Added a new dummy_sig method for XSpectrum1D to generate an 
uncertainty array in specutils' format.  A different name might be preferred.


linetools/spectra/xspectrum1d.py
```diff
@@ -133,6 +133,17 @@ def add_noise(self,seed=None,s2n=None):
         #
         self.flux = self.flux + (rand * sig)*self.flux.unit
 
+    #  Add a dummy sigma array via uncertainty
+    def dummy_sig(self,sigv=0.):
```
ntejos: How about calling this method 'uniform_sig()'?

nhmc:
Scripts like this are great, let's definitely put it in linetools.  But it should follow the astropy format: http://docs.astropy.org/en/stable/development/scripts.html

Doing that will also make it easier to test, because you can just import the module.

We should also delay as many imports as possible until after parser.parse_args(args) is called. This makes the script fast if it's just printing an input error or help message.

profxj:
Ah, now you show me this! 

Conformed to astropy and added some documentation.

linetools/scripts/linet_absline.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
```
nhmc: This line can be removed

linetools/scripts/linet_absline.py
```diff
@@ -0,0 +1,67 @@
...
+  linet_absline 1334.3    13.5 7
+  linet_absline 1215.6701 14.0 30
+"""
+
+import pdb
+
+
+def plot_absline(wrest,logN,b):
+    """Plot an absorption line with N,b properties
+    Parmaeters
```
nhmc: typo

nhmc:
Thanks. Can we use `lt_` as a script prefix instead of `linet_`?

profxj:
so demanding...

Now it is so.

docs too.

nhmc:
Awesome, thanks :) Please merge when ready.

profxj:
Passed tests.  
Merging.

PR #80, opened 2015-11-18
Installation docs

nhmc:
A doc page on installation.  I haven't actually checked that following these instructions works...


profxj:
Nice!

I hadn't realized we were 'pip-ified'! 

I'll have someone test this shortly.

X

PR #82, opened 2015-11-18
XSpectrum1D diagnostics

profxj:
Performs simple diagnostics on the spectrum.

Might running this as a default and putting in as attributes.
Thoughts?

Added a test too.  Would have added documentation, but
we aren't there yet. ;)


ntejos:
I like the idea to have this kind of information as attributes.

profxj:
That always exist, i.e. are generated when instantiated?

Just checking.

X

----------------------------------------------
J. Xavier Prochaska
UCO/Lick Observatory
1156 High St.
UC Santa Cruz
Santa Cruz, CA 95064

http://www.ucolick.org/~xavier/
http://imps.ucolick.org/
http://www.qpqsurvey.org/
http://k1dm3.ucolick.org/
http://www.ucolick.org/~xavier/IDL/index.html
https://github.com/profxj/xastropy

831-459-2135 (Direct)
831-459-2991 (UCO/Lick Main)
831-459-5244 (Fax)

profxj:
Actually, should I use this type of approach?

    @property
    def sig(self):
        ''' Return a standard 1sigma error array
        '''
        if isinstance(self.uncertainty,StdDevUncertainty):
            return self.uncertainty.array
        else:
            return None

ntejos:
Yep. I'd like a XSpectrum1D object to have a minimum of attributes that
make it to be a sensible spectrum. Regarding the way to initialize them, I
like what you propose; looks clean. But maybe Neil has a better idea.

---Nicolas

linetools/spectra/xspectrum1d.py
```diff
@@ -144,6 +143,37 @@ def constant_sig(self,sigv=0.):
...
+            median S/N
+        """
+        # Cut on good pixels
+        gdpx = self.sig > 0.
+        # Generate
+        diag_dict = dict(
+            wave_min=np.min(self.dispersion[gdpx]),
+            wave_max=np.max(self.dispersion[gdpx]),
+            med_flux=np.median(self.flux[gdpx]),
+            med_s2n=np.median(self.flux[gdpx]/self.sig[gdpx]).value,
```
nhmc: np.median(self.flux[gdpx].value/self.sig[gdpx].value)

to avoid unit conversions

nhmc:
This looks good to me.  You could print this info in the __repr__ method so it will show up in the interpreter. e.g. something like

```
>>> spec
XSpectrum1D:  wmin=XX, wmax=XX, median flux=XX, median S/N=XX 
```
I also like them as attributes. Maybe median_s2n, median_flux, wmin, wmax?

Let me think more about @property...

profxj:
I dont think I can fill up the attributes when it is instantiated because
I am using the Spectrum1D instantiation and don't wish to monkey with that code.

So, if we want these to always be accessible, they'll need to be as
something like a property.  I think.

profxj:
I take that back.

Stand by.

nhmc:
Now that I think about it, using @property would indeed be better, because it only runs set_diagnostics when needed ('lazy' evaluation). The approach I've seen used is:

```
    @property
    def wmin(self):
        ''' Minimum wavelength '''
        if self._wmin is None:
            # this set_diagnostics sets the attributes _wmin, _wmax, etc.
            self.set_diagnostics()
        else:
            return _wmin
```

profxj:
Ok, at it again.

Decided to go with only wvmin and wvmax for now
(wasn't biting on wmin).

X

nhmc:
Merging

PR #83, opened 2015-11-20
Update spectralline.py so Sphinx can parse docstrings

nhmc:
This updates `spectralline.py` so that Sphinx will parse the docstrings correctly (according to the rules given at http://docs.astropy.org/en/stable/development/docrules.html). I suggest we use `spectralline.py` as a template for future docstrings.

You can see the result here: https://dl.dropboxusercontent.com/u/7970232/linetools_docs/index.html
(go to linetools API -> SpectralLine, for example).

I've also updated the installation docs with instructions on how to build the docs.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,9 +1,8 @@
-================
+=================
 The linetools API
-================
+=================
 
-.. automodule:: spectralline
-    :members:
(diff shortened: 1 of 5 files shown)


profxj:
Wow, super impressive Neil.  
Thanks for really digging into this.

Merging

PR #87, opened 2015-11-24
Add a copy method to XSpectrum1D

nhmc:
This adds a copy method to XSpectrum1D, and allows a continuum input to XSpectrum1D.from_tuple.

There are also a few doc tweaks and minor bugfixes (sorry to lump them together).


profxj:
Changes look fine but tests are failing.

Fix and then merge.

PR #93, opened 2015-11-27
Make linetools/spectra docstrings work with Sphinx

nhmc:
Also remove:
 - the `spectra/utils.py` file, which appears to do nothing
 - `XSpectrum1D.parse_spec`, which isn't used anywhere. We should be probably be using `Spectrum1D.slice_index` or `.slice_dispersion` anyway.

Thanks to @ntejos for the nice docstrings in lsf and convolve :)


nhmc:
Merging

PR #94, opened 2015-11-27
Finish cleaning up docstrings

nhmc:
All the docstrings should now be pretty much done. Thanks for the recent cleaning up @profxj, it makes things much easier!


profxj:
Well done, @nhmc!

Merging

PR #96, opened 2015-11-29
AbsSurvey -- Beginnings

profxj:
AbsSurvey class

Pretty bare bones but includes:
test
docs 
notebook

### DIFF
--- a/linetools/isgm/abssurvey.py
+++ b/linetools/isgm/abssurvey.py
@@ -14,6 +14,7 @@ from astropy.table import QTable, Column
 from astropy.units.quantity import Quantity
 
 from linetools.spectra import io as lsio
+from linetools.isgm.abssystem import AbsSystem
 
 #
 
@@ -73,6 +74,11 @@ class AbslineSurvey(object):
         """Generate the Survey from a summary FITS file
 
         Handles SPEC_FILES too.
+
+        Parameters
+        ----------
+        summ_fits : str
+          Summary FITS file
         """
         # Read
         systems = QTable.read(summ_fits)
@@ -117,7 +123,7 @@ class AbslineSurvey(object):
         # Return
         return slf
 
-    def __init__(self, abs_type, nsys=0, ref=''):
+    def __init__(self, abs_type, ref=''):
         # Expecting a list of files describing the absorption systems
         """  Initiator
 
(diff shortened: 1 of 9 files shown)


linetools/isgm/abssurvey.py
```diff
@@ -0,0 +1,325 @@
...
+        abs_type : str
+          Type of AbsSystem in the Survey, e.g.  MgII, DLA, LLS
+        nsys : int, optional
+          Number of AbsSystem in the Survey
+        ref : string, optional
+          Reference(s) for the survey
+        """
+        self.abs_type = abs_type
+        self.ref = ref
+        self._abs_sys = []
```
nhmc: Why does this have a leading underscore?

linetools/isgm/abssurvey.py
```diff
@@ -0,0 +1,325 @@
...
+                    t.add_row( row )
+                continue
+            else:
+                raise ValueError('Multiple entries...')
+
+        # Return
+        return t[1:]
+
+    # Mask
+    def upd_mask(self, msk, increment=False):
```
nhmc: I prefer `update_mask`

linetools/isgm/abssurvey.py
```diff
@@ -0,0 +1,325 @@
+""" Class for and AbslineSurvey
+"""
+
+from __future__ import print_function, absolute_import, division, unicode_literals
+
+import numpy as np
+import imp, json, copy
+from abc import ABCMeta
```
nhmc: This import can be removed (also please check for any other unused imports)

nhmc: Oops, I now see it is used AbslinSurvey. Disregard :)

linetools/isgm/abssurvey.py
```diff
@@ -0,0 +1,325 @@
...
+    """Translate abstype into Class
+    Parameters:
+    ----------
+    abstype: str
+      AbsSystem type, e.g. 'LLS', 'DLA'
+
+    Returns:
+    --------
+    Class name
+    """
```
nhmc: Please make docstrings sphinx-friendly.

nhmc:
I left some inline comments.  Also, it seems like nsys has to be specified in addition to adding each system? If so it would be much better to make nsys a `@property` that returns `len(_abs_sys)` when called.

linetools/isgm/abssurvey.py
profxj: It is very rare that one will access this list of objects directly.

I could, however, name it _abssys if that is easier to read.

linetools/isgm/abssurvey.py
profxj: Done

linetools/isgm/abssurvey.py
nhmc: But in the example notebook you access it directly to add new systems. If that's not the intended usage, the example should be changed.  (Also `_abs_sys`  is way better than `_abssys`)

profxj:
I like the property idea and have put it in.

I may have had it in mind to allows a Survey without filling
the abs_sys list but nevermind that.

nhmc:
I see you fixed the docs too. Please merge when ready.

profxj:
Merging

PR #99, opened 2015-12-02
Abs component

ntejos:
Added self.name attribute to AbsComponent
Added self.comment attribute to AbsComponent
Added repr_vpfit() to AbsComponent
Other minor stuff

### DIFF
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -134,7 +134,7 @@ class AbsComponent(object):
 
         # Required
         if isinstance(radec,(tuple)):
-            self.coord = SkyCoord(ra=radec[0], dec=radec[1])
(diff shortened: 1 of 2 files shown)


linetools/isgm/abscomponent.py
```diff
@@ -411,6 +422,32 @@ def stack_plot(self, nrow=6, show=True):
             plt.show()
         plt.close()
 
+
+    def repr_vpfit(self, b=10.):
```
nhmc: Please add in the docstring whether this is fort.26 or fort.13 format. I always get confused between the two.

nhmc:
Can you add a test for repr_vpfit? After that, feel free to merge.

linetools/isgm/abscomponent.py
ntejos: Done.

PR #101, opened 2015-12-02
DLA: Subclass for AbsSystem and AbsSurvey

profxj:
Found I had to touch a number of other things.

tests
and docs too


linetools/isgm/dla.py
```diff
@@ -0,0 +1,247 @@
...
+          Defaulted to +/- 500 km/s if None
+        NHI : float, required despite being a keyword
+          log10 of HI column density
+          must be 20.3 or higher
+        **kwargs : keywords
+          passed to AbsSystem.__init__
+        """
+        # NHI
+        try:
+            NHI = kwargs['NHI']
```
nhmc: replace this with `NHI = kwargs.pop('NHI')` and remove the else: block below

linetools/isgm/dla.py
```diff
@@ -0,0 +1,247 @@
...
+        slf.flg_ZH = float(slf._datdict['flg_mtl'])
+        slf.ZH = float(slf._datdict['[M/H]'])
+        slf.sig_ZH = float(slf._datdict['sig([M/H])'])
+
+        return slf
+
+    def __init__(self, radec, zabs, vlim, **kwargs):
+        """Standard init
+
+        NHI keyword is required
```
nhmc: Why isn't it a positional parameter like to others, then? You can still pass positional parameters as keywords if you want to do this just to be explicit. e.g.

``` python
def f(a,b): print(a, b)
f(1, 2)
f(b=2, a=1)
```
do the same thing

linetools/isgm/dla.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Parameters
+        ----------
```
nhmc: Move the parameter description to the class docstring for Sphinx.

linetools/isgm/dla.py
nhmc: (if you keep it as a keyword)

linetools/isgm/dla.py
```diff
@@ -0,0 +1,247 @@
...
+        idict : dict, optional
+          dict containing the IonClms info
+        use_clmfile : bool, optional
+          Parse ions from a .clm file (JXP historical)
+        update_zvlim : bool, optional
+          Update zvlim from lines in .clm (as applicable)
+        linelist : LineList
+        """
+        reload(ltiu)
```
nhmc: Remove left over reloads

linetools/isgm/dla.py
```diff
@@ -0,0 +1,247 @@
...
+            self._ionN = ltiu.iontable_from_components(components, ztbl=self.zabs)
+            # Add to AbsSystem
+            for comp in components:
+                self.add_component(comp)
+        else:
+            raise IOError("Not ready for this")
+
+    # Output
+    def __repr__(self):
+        return ('[{:s}: {:s} {:s}, {:g}, NHI={:g}, Z/H={:g}]'.format(
```
nhmc: I think we should use '< >' instead of '[ ]' for repr strings, because '[ ]' makes it look like a list.

linetools/isgm/abssurvey.py
```diff
@@ -253,7 +251,7 @@ def ions(self, iZion, skip_null=False):
         Table of values for the Survey
         """
         keys = [u'name', ] + self.abs_sys()[0]._ionN.keys()
-        t = copy.deepcopy(self.abs_sys()[0]._ionN[0:1])
+        t = Table(copy.deepcopy(self.abs_sys()[0]._ionN[0:1]))  # Avoids mixin trouble
```
nhmc: I never really know what is going on with deepcopy and classes (!). It would be nice to make a `copy` method for any class object we end up needing deepcopy for.

linetools/isgm/abssurvey.py
```diff
@@ -212,7 +212,7 @@ def __getattr__(self, k):
         # Recast as an array
         return lst_to_array(lst, mask=self.mask)
 
-    def fill_ions(self, jfile=None):  # This may be overloaded!
+    def fill_ions(self, use_clmfile=False,jfile=None):  # This may be overloaded!
```
nhmc: I would prefer `use_Nfile` to `use_clmfile`.

linetools/isgm/abscomponent.py
```diff
@@ -460,20 +459,14 @@ def __getitem__(self, attrib):
...
 
     def __repr__(self):
-        txt = '[{:s}: {:s} {:s}, Name={}, Zion=({:d},{:d}), z={:g}, vlim={:g},{:g}'.format(
-            self.__class__.__name__,
-            self.coord.ra.to_string(unit=u.hour,sep=':',pad=True),
-            self.coord.dec.to_string(sep=':',pad=True,alwayssign=True),
-            self.name,
-            self.Zion[0], self.Zion[1], self.zcomp,
-            self.vlim[0],self.vlim[1])
+        txt = '[{:s}: {:s} {:s}, Name={:s}, Zion=({:d},{:d}), Ej={:g}, z={:g}, vlim={:g},{:g}'.format(
```
nhmc: I prefer '< >' to '[ ]' in repr (see below).

linetools/isgm/dla.py
```diff
@@ -0,0 +1,247 @@
...
+                self.__class__.__name__,
+                 self.coord.ra.to_string(unit=u.hour, sep=':', pad=True),
+                 self.coord.dec.to_string(sep=':', pad=True),
+                 self.zabs, self.NHI, self.ZH))
+
+    def print_abs_type(self):
+        """"Return a string representing the type of vehicle this is."""
+        return 'DLA'
```
nhmc: Wouldn't this better as just an attribute called `abs_type`? Seems weird to make it a method.

(I know it's also in the base class GenericAbsorptionSystem, so it may be a bit late to be bringing this up...)

linetools/isgm/dla.py
profxj: Done. Thx.

linetools/isgm/dla.py
profxj: Good point.  Now positional.

linetools/isgm/dla.py
profxj: Done.

linetools/isgm/dla.py
profxj: Gone

linetools/isgm/dla.py
profxj: I agree, although [] seems to be the standard.  

Anyways, have switched it.

linetools/isgm/abssurvey.py
profxj: Well, this is just copying a snipped of an astropy Table to
get the rows.  Maybe there is a method in that already..

linetools/isgm/abssurvey.py
profxj: ok

linetools/isgm/abscomponent.py
profxj: done.

linetools/isgm/dla.py
profxj: Leaving it for now.

linetools/isgm/abssurvey.py
nhmc: Then I believe `self.abs_sys()[0]._ionN[0:1].copy()` should work.

nhmc:
Thanks! After looking at whether you can remove deepcopy, merge when ready.

linetools/isgm/abssurvey.py
profxj: Ok, I thikn I got it.

profxj:
merging.

PR #102, opened 2015-12-02
More doc changes

nhmc:
Now down to just 10 Sphinx warnings (yay!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,20 @@
 language: python
 
+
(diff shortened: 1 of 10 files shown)


nhmc:
I tried adding a doc build to the Travis tests, but this produced warnings that I can't reproduce building on my machine. So the doc build test is still disabled for now.

nhmc:
Merging.

PR #103, opened 2015-12-03
 LLS: Some edits for the LLS GUI

profxj:
A new method or two
and a test

### DIFF
--- a/linetools/isgm/abssystem.py
+++ b/linetools/isgm/abssystem.py
@@ -172,10 +172,31 @@ class AbsSystem(object):
         """Additional checks on the component"""
         return True
 
+    def list_of_abslines(self):
+        """ Generate a list of the absorption lines in this system
+
+        Drawn from the components
+
+        Returns
+        -------
+        abslist : list
+
+        """
+        abslist = []
+        for component in self._components:
+            for iline in component._abslines:
+                abslist.append(iline)
+        # Return
+        return abslist
+
+
 class GenericAbsSystem(AbsSystem):
     """Class for Generic Absorption Line System
     """
     def __init__(self, radec, zabs, vlim, **kwargs):
+        if vlim is None:
+            if vlim is None:
+                vlim = [-500.,500.]*u.km/u.s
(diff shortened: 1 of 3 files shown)


linetools/isgm/abssystem.py
```diff
@@ -172,10 +172,31 @@ def chk_component(self, component):
...
+        # Return
+        return abslist
+
+
 class GenericAbsSystem(AbsSystem):
     """Class for Generic Absorption Line System
     """
     def __init__(self, radec, zabs, vlim, **kwargs):
+        if vlim is None:
+            if vlim is None:
```
ntejos: Duplicate statements.

profxj:
Fixed.

Anything else?

linetools/isgm/abssystem.py
```diff
@@ -172,10 +172,30 @@ def chk_component(self, component):
...
+        # Return
+        return abslist
+
+
 class GenericAbsSystem(AbsSystem):
     """Class for Generic Absorption Line System
     """
     def __init__(self, radec, zabs, vlim, **kwargs):
+        if vlim is None:
+            vlim = [-500.,500.]*u.km/u.s
```
ntejos: I'm still thinking whether this is a good idea or not. If you do something like this for vlim, then for the sake of symmetry I would also allow radec and zabs to have default values, which may be useful in some cases where you do not care about those quantities. On the other hand, for the cases you do care, having default values may cause some trouble...

ntejos: How about having them as optional parameters?

linetools/isgm/lls.py
```diff
@@ -14,6 +14,7 @@
 from linetools.analysis import absline as ltaa
```
ntejos: Should you move lls.py to the IGM package instead?

linetools/isgm/abssystem.py
profxj: Yeah, fair point.

I'll go back to what we had.

linetools/isgm/lls.py
profxj: Not today and this needs to be done now.

PR #106, opened 2015-12-06
Refactoring IGM Systems

profxj:
Mainly removed the systems (LLS, DLA)
And the Surveys
and the docs

All of this is now in pyigm


nhmc:
Merging

PR #112, opened 2015-12-10
Add docs for the SolarAbund class

nhmc:
Also some other minor changes.

### DIFF
--- a/linetools/abund/elements.py
+++ b/linetools/abund/elements.py
@@ -66,7 +66,8 @@ Examples
 
 """
 
-from __future__ import division, print_function
+from __future__ import (division, print_function, unicode_literals,
+                        absolute_import)
 
 __version__ = '2015.01.29'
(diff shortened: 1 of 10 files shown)


linetools/abund/elements.py
```diff
@@ -290,7 +291,7 @@ def __str__(self):
                                      self.abundance * 100)
 
     def __repr__(self):
-        return "Isotope(%s, %s, %s)" % (
+        return "<Isotope(%s, %s, %s)>" % (
```
profxj: maybe change to {:s} format, if you can

profxj:
Looks good.  Pass the test and merge.

linetools/abund/elements.py
nhmc: There are 30+ strings to do in that module, so I might skip this for now.

nhmc:
Merging.

PR #114, opened 2015-12-11
Remove convolve_window, doc changes, normalise gaussian smoothing kernel

nhmc:
This has a few changes (sorry):

 * Remove the convolve_window function, which I don't think is used anywhere
 * Remove the rest highlighting commands at the top of the doc files, because they prevent code highlighting in examples
 * Normalise the Gaussian smoothing kernel in convolve_psf. Without this,  a spectrum's normalisation changes a lot when you smooth it with a narrow psfs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 
-
 # The apt packages below are needed for sphinx builds, which can no longer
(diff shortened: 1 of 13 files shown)


profxj:
Looks good.

Merging..

PR #116, opened 2015-12-13
xspectrum1d: Exten keyword for multi-extension FITS

profxj:
only useful for binary tables, e.g. PYPIT


profxj:
need this now.

merging.

PR #119, opened 2015-12-18
LineList: Galaxy line list

profxj:
Emission and absorption lines for galaxy spectra
Data is mainly limited to wavelength

docs, tests


nhmc:
Maybe add a `ref` field for the galaxy lines, similar to `Verner94`, etc. for the ISM lines? Otherwise it looks great.

profxj:
Ok, I've added something.  
There is no paper to cite, I fear.
And the absorption lines are especially uncertain.  I really couldn't 
get *anyone* to give me a list so I used a very old one that I had gotten
from Dan Kelson (origin otherwise unknown).

profxj:
merging

PR #123, opened 2015-12-21
Remove Pyqt import

nhmc:
This was making the readthedocs build fail.


nhmc:
Merging. This just removes an import statement, so the test coverage decrease is not an issue.

PR #124, opened 2015-12-22
Remove Akima1DInterpolator import

nhmc:
because it's not used, and makes us incompatible with scipy versions < 0.14 (apparently used by readthedocs...)


nhmc:
Merging.

PR #126, opened 2015-12-22
Xspectrum1d

ntejos:
Added some functions to deal with continuum in Xspectrum1d:
- perturb_continuum()
- reset_continuum()
- _interp_continuum()

I thought about adding these in the continuum module, but it looked more natural to have them implemented in the Xspectrum1d class, given that there are already functions for adding noise and convolving... Please let me know if you think otherwise.


linetools/spectra/xspectrum1d.py
```diff
@@ -236,49 +239,49 @@ def constant_sig(self, sigv=0.):
...
-                raise ValueError('normalize: Continuum needs to be same length as flux array')
+                raise ValueError(
+                    'normalize: Continuum needs to be same length as flux array')
 
         # Adjust the flux
-        self.flux /= conti
+        self.flux /= co
         if self.uncertainty is not None:
-            self.uncertainty.array /= conti
+            self.uncertainty.array /= co
```
profxj: I prefer conti to co myself, but I'm going
to let it go..

profxj:
I expect that these will end up outside of XSpectrum eventually,
but I'm ok with it for now.

I am making a last vote for co instead of conti but I can live with co..

I leave it to Neal to merge.

linetools/spectra/xspectrum1d.py
```diff
@@ -772,12 +781,106 @@ def fit_continuum(self, knots=None, edges=None, wlim=None, dw=10.,
...
+        xy = np.array(contpoints)
+        xy = xy.transpose()
+        x, y = xy[0], xy[1]
+
+        # Get indices in the full spectrum
+        # inds = np.array([np.where(np.fabs(x[i]-self.dispersion.value)<0.01)[0][0] for i in range(len(x))])
+        # y_err = y_err = self.uncertainty.array[inds]
+
+        # Seed
+        np.random.seed(seed=seed)
```
nhmc: I would only call random.seed here if seed is not `None`. This allows you to set the seed at the beginning of a script and then call this method many times in a reproducible sequence without having to give a seed every time.

Incidentally, passing around a RandomState instance is better way to make a reproducible set of random numbers than setting the seed. See Robert Kern's answer to this question: http://stackoverflow.com/questions/5836335/consistenly-create-same-random-numpy-array (Joe Hennawi discovered this). But we can add this later if necessary.

linetools/spectra/xspectrum1d.py
```diff
@@ -772,12 +781,106 @@ def fit_continuum(self, knots=None, edges=None, wlim=None, dw=10.,
...
+        #add Normal noise to y points
+        y += np.random.normal(0, y * rel_var, len(y))
```
nhmc: I think this line doesn't modify the original meta['contpoints'] because it's copied by the `xy = np.array(contpoints)` above. But can you change this `y = y + np.random.normal(...` just make sure?

linetools/analysis/continuum.py
```diff
@@ -95,6 +95,27 @@ def update_knots(knots, indices, fl, masked):
...
+       The start and end indices into fl and masked of each
+       spectrum chunk (xpos of each knot are the chunk centres).
+    sig_fl, masked : arrays shape (M,)
+       The error in the flux, and boolean arrays showing which
+       pixels are masked.
+    """
+    sig_y = []
+    pass
+
+
```
nhmc: Can you remove this? If you want to keep this code around in case we want to add a full version later, you could make a branch in your repo.

nhmc:
Thanks!  I left some comments. Feel free to merge after looking at those.

linetools/spectra/xspectrum1d.py
ntejos: Actually, my intention was not to change the original `meta[contpoints]` so we can reset to the original eventually...

linetools/analysis/continuum.py
ntejos: Sure, will do.

linetools/spectra/xspectrum1d.py
ntejos: Got it. Not sure on how to implement passing the RandomState() though... will leave it for later. There are other places in linetools where pseudo random numbers are called, we should probably address all of them in a similar fashion (if possible).

PR #127, opened 2015-12-24
Fix bug in continuum fitting

nhmc:
This fixes a bug in the continuum fitting code: it would fail to save continuum points if the spectrum's flux is type numpy.float32, instead of the numpy float64 (which is the native python type, I think).

It also adds a (slightly) more informative error message if no continuum spline points are found, and cleans up the loadjson and savejson functions.


profxj:
Thanks for this.  Will be happy to use it.

PR #128, opened 2015-12-24
AbsSystem: Tools for IGMSurvey (in pyigm)

profxj:
pushed the radec to coord code to linetools.utils
added a skip_vel option in AbsComponent.add_absline

new tests for the radec


linetools/isgm/abscomponent.py
```diff
@@ -157,7 +157,7 @@ def __init__(self, radec, Zion, z, vlim, Ej=0./u.cm, A=None, Ntup=None, comment=
         self.name = ions.ion_name(self.Zion, nspace=1)
         self._abslines = []
 
-    def add_absline(self, absline, tol=0.1*u.arcsec):
+    def add_absline(self, absline, tol=0.1*u.arcsec, skip_vel=False):
```
ntejos: This seems very risky to me; please add a warning message. Wouldn't it be better to have a velocity tolerance which by defaults is that of the Component itself, but that can also be set to arbitrary large values or None (which is effectively a skip_vel = True).

ntejos:
Left a inline comment. The rest look good to me so please feel free to merge after considering the comment.

linetools/utils.py
```diff
@@ -34,6 +34,47 @@ def between(a, vmin, vmax):
     c &= a >= vmin
     return c
 
+def radec_to_coord(radec):
+    """ Convert one of many radec formats to SkyCoord
+
+    Parameters
+    ----------
+    radec : str or tuple
```
nhmc: Please list some examples of valid strings e.g. 'SDSSJ010101+010101'

nhmc:
Apart from my inline comment this looks good to me. I'm agnostic about @ntejos' skipvel suggestion, do what you think is best.

linetools/utils.py
profxj: added.

linetools/isgm/abscomponent.py
profxj: I prefer it the way it is.  
One has to purposefully choose skip_vel to 
use it.  And otherwise the abs_line is required
to fall within the Component.

profxj:
merging

PR #131, opened 2015-12-28
LineList gal

ntejos:
- Unify `Galaxy` and `ISM` LineLists, so all have the same `_fulltable` format with masked values when needed.
- Added reference to EUV lines by hand pointing to `Tripp`
- Other minor tweaks


linetools/lists/linelist.py
```diff
@@ -40,6 +44,7 @@ class LineList(object):
        * 'H2'      :: H2 (Lyman-Werner)
        * 'CO'      :: CO UV band-heads
        * 'EUV'     :: EUV lines (for CASBAH project)
+       * 'Galaxy'  :: Lines typically identified in galaxy spectra
        * 'Gal_E'   :: Galaxy emission lines (HII, to be implemented)
```
profxj: Maybe we should eliminate Gal_E and Gal_A ?

ntejos: I agree. Will do so.

profxj:
I made one comment about Gal_E and Gal_A.

Stew on that and then merge.

Thanks!

PR #132, opened 2015-12-29
Docs: LineList

ntejos:
- Added documentation in linelist.rst
- Solved an issue with subset sorting in linelist.py


profxj:
Looks fine.

Meriging

PR #135, opened 2015-12-31
xspectrum1d: Rebinning error array

profxj:
Modified the rebin method to handle the error array.
S/N is conserved in a kludgy manner
test + docs


nhmc:
This looks good to me. After the conflicts are resolved, feel free to merge.

In the future it would be nice to give more information about rebinning in the docs. We could mention that it introduces correlations between neighbouring pixels, for example, and point to literature about different ways to rebin.

profxj:
Resolved the conflicts and added some (brief) 
documentation on neighboring pixels.  Wasn't
sure what literature to cite..

profxj:
Merging..

PR #140, opened 2016-01-07
dicts: AbsSystem and AbsComponent

profxj:
read and write methods using dict's for
AbsSystem and AbsComponent
docs
tests


nhmc:
Looks good. Merging.

PR #141, opened 2016-01-11
AbsSystem:  More dict

profxj:
A bit more related to read/write of AbsSystem, components, etc.


profxj:
Passed tests.
Merging

PR #154, opened 2016-01-20
xspec gui enhancements

nhmc:
This makes some changes to the xpsec gui to make the startup time faster. It 

 - removes a redundant call to guis.utils.read_spec 
 - shifts around some imports
 - replaces try-except statements, because catching exceptions is slow.

My subjective impression is that these speed up startup by a bit.


profxj:
All good.
Merging.

PR #163, opened 2016-01-25
 LineList: Making subset_lines() to return a LineList object instead.

ntejos:
LineList: Making subset_lines() to return a LineList object instead.

Initial commit; will use Travis to check up crashes from tests and will fix them one at a time.

### DIFF
--- a/docs/linelist.rst
+++ b/docs/linelist.rst
@@ -134,7 +134,7 @@ original LineList object. Consider that for some reason you may want
 only HI Lya and Lyb in your LineList, then you can achieve this by::
 
   hi = LineList('HI')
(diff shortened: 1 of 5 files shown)


ntejos:
The current failure seems to do with copy.copy() in python 3, but seems to be working fine in python 2.7 ... any thoughts?

nhmc:
I have a fix - I will merge this and then apply it.

nhmc:
Actually, I can't merge because there are conflicts. Could you resolve the conflicts, and then apply the changes in the attached diff?

[diff.txt](https://github.com/linetools/linetools/files/107620/diff.txt)

ntejos:
Sure, will do.

nhmc:
Thanks. Merging.

PR #167, opened 2016-01-27
Add a link to the original examples notebooks in the sphinx rst versions

nhmc:
This alters the notebook conversion script so that at the beginning of each generated .rst file there's a link to the original notebook file.


nhmc:
Merging.

PR #169, opened 2016-01-27
Documentation

profxj:
Mainly SpectralLine and AbsLine


profxj:
hopefully that worked..

PR #172, opened 2016-01-28
Fix lt_absline, other bug fixes

nhmc:
This fixes the lt_absline script and adds a test. It also adds an option to set wavelength limits when initiating an Interactive plot (also for ExamineSpec widget).


nhmc:
Merging.

PR #174, opened 2016-01-29
Add table to speed up reading ISM-related line data

nhmc:
Reading the ISM linelist and related versions (Strong, EUV)  is very slow. This PR fixes this by 

 1. Adding a new function in linelist.py to write out ISM data as a fits file, together with columns giving the Strong, EUV and HI subsets.
 2. Reading this file instead of all the source files by default when making a new list. This can be turned off by the `use_ISM_table` keyword to LineList.

I also removed the possibility of making a line list with multiple keys (e.g. LineList(['ISM', 'molecules'])) because this (a) simplifies the code, (b) was broken for some molecule combinations and (c) seems to be of limited use.

Also a test is added. Nothing should change from the user point of view, apart from multiple keys not working anymore.


nhmc:
@profxj @ntejos Could you take a look at this if you're still working?

linetools/lists/linelist.py
```diff
@@ -718,10 +723,51 @@ def __getitem__(self, k, tol=1e-3 * u.AA):
...
     # Printing
     def __repr__(self):
-        # Generate sets string
-        for kk, llist in enumerate(self.lists):
-            if kk == 0:
-                sstr = llist
+        return '<LineList: {:s}; {} transitions>'.format(self.list, len(self._data))
+
+
+def _write_ref_ISM_table():
```
profxj: You might shove this into parse.py instead.

profxj:
One minor comment from me.

Can you get it to pass the tests??

If so, merge.  If not and you feel ok about it then merge.  :)

nhmc:
Thanks!

To make the tests pass I had to make jsonify_dict deal with booleans and tuples. So it's now replaced by jsonify, which is no longer dictionary-specific.

ntejos:
Great! Could you also update the linelist.rst to account for the changes?

ntejos:
I'm happy to update the documentation myself, but it would need to be tomorrow morning PST...

PR #175, opened 2016-01-29
Continuum fitting GUI fix, doc changes

nhmc:
This fixes the continuum fitting GUI, updates the notebooks again, and fixes the linelist documentation.

Sorry for lumping so many changes into one PR!


ntejos:
BTW Neil, what happened with the capability of adding a break (i.e. discontinuity) in the continuum fitting GUI?

nhmc:
You should be able to just add a few knots close together and it will approximate a break. The ability to add a discontinuity was originally there because putting knots too close together using standard cubic splines resulting in 'ringing'. But Akima splines don't have than problem, so I took it out.

nhmc:
Merging.

ntejos:
I see, thanks for the clarification.

PR #176, opened 2016-01-29
Xspectrum1d: normalize/unnormalize

ntejos:
- I've added a method for un-normalizing a spectrum.
- Added boolean attribute ``self.normed`` which helps in avoiding the possibility of `normalize` a normalized spectrum (i.e. divide by the continuum again).


profxj:
Merging.

PR #181, opened 2016-01-30
Docs edits

profxj:
removed pyaml as a dependency in install
changed pip call too
small edits to linelist.rst


nhmc:
Looks good, merging.

PR #183, opened 2016-02-01
Script for Continuum Fit

profxj:
new script
some docs
a few other edits

### DIFF
--- a/docs/scripts.rst
+++ b/docs/scripts.rst
@@ -66,3 +66,14 @@ Try::
 	lt_absline -h
 
 for the full set of options.
+
+lt_continuumfit
+---------------
+
+Launch the GUI to continuum fit a spectrum.
+If a redshift is supplied by zsys, then the
+script assume this is a QSO.::
+
+   lt_continuumfit fitsfile.fits
+
+
--- a/docs/scripts.rst
+++ b/docs/scripts.rst
@@ -72,7 +72,7 @@ lt_continuumfit
 
 Launch the GUI to continuum fit a spectrum.
 If a redshift is supplied by zsys, then the
-script assume this is a QSO.::
+script assumes this is a QSO.::
 
    lt_continuumfit fitsfile.fits
 
--- a/linetools/guis/utils.py
+++ b/linetools/guis/utils.py
@@ -214,7 +214,10 @@ def read_spec(ispec, exten=None, norm=True):
 
     # Normalize?
     if norm:
-        spec.normalize()
+        try:
+            spec.normalize()
+        except ValueError:
+            norm = False
 
-    # Return
+   t# Return
     return spec, spec_fil
--- a/linetools/guis/utils.py
+++ b/linetools/guis/utils.py
@@ -214,7 +214,7 @@ def read_spec(ispec, exten=None, norm=True):
 
     # Normalize?
     if norm:
-        if spec.co is not None:
+        if hasattr(spec, 'co') and spec.co is not None:
             spec.normalize()
 
     # Return
--- a/linetools/guis/utils.py
+++ b/linetools/guis/utils.py
@@ -214,10 +214,8 @@ def read_spec(ispec, exten=None, norm=True):
 
     # Normalize?
     if norm:
-        try:
+        if spec.co is not None:
             spec.normalize()
-        except ValueError:
-            norm = False
 
     # Return
     return spec, spec_fil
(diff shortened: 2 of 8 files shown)


docs/scripts.rst
```diff
@@ -66,3 +66,14 @@ Try::
 	lt_absline -h
 
 for the full set of options.
+
+lt_continuumfit
+---------------
+
+Launch the GUI to continuum fit a spectrum.
+If a redshift is supplied by zsys, then the
+script assume this is a QSO.::
```
nhmc: typo 'assumes'

linetools/guis/utils.py
```diff
@@ -212,5 +212,10 @@ def read_spec(ispec, exten=None):
     else:
         raise ValueError('Bad input to read_spec: {:s}'.format(type(ispec)))
 
+    # Normalize?
+    if norm:
+        if spec.co is not None:
```
nhmc: Add `if hasattr(spec, 'co') and...`. I don't think we guarantee the co attribute exists (maybe we should).

linetools/scripts/lt_continuumfit.py
```diff
@@ -0,0 +1,46 @@
...
+import sys
+
+
+# Script to run continuum fitting in linetools
+def main(*args, **kwargs):
+    """ Runs the continuum fitter
+    """
+    import argparse
+
+    parser = argparse.ArgumentParser(description='Parse for Continuum Fitting')
```
nhmc: The description string is what shows up when you call 'lt_continuum -h'. So it should say what the script does, i.e. "Fit a continuum to a spectrum".

linetools/scripts/lt_continuumfit.py
```diff
@@ -0,0 +1,46 @@
...
+    parser.add_argument("file", type=str, help="Spectral file")
+    parser.add_argument("outfil", type=str, help="Output file [can be the same]")
+    parser.add_argument("-zsys", type=float, help="Redshift of the Source")
```
nhmc: Maybe change this to ('-z', '--redshift', ... ) instead of just '-zsys'?

linetools/scripts/lt_continuumfit.py
```diff
@@ -0,0 +1,46 @@
...
+    parser.add_argument("-wchunk", type=float, help="Width of a 'chunk' (Ang)")
```
nhmc: I suggest we use the convention that a single '-' should have a single character flag (i.e. '-h'), and the double '--' should be used for multiple character flags. So these would be changed to '--zsys' and `--wchunk`.

linetools/scripts/lt_xspec.py
```diff
@@ -16,7 +16,7 @@ def main(*args, **kwargs):
     parser = argparse.ArgumentParser(description='Parse for XSpec')
     parser.add_argument("file", type=str, help="Spectral file")
     parser.add_argument("-zsys", type=float, help="System Redshift")
```
nhmc: see comment above about using '--' for multiple character flags

nhmc:
Thanks, this looks great!  Please merge after considering the comments.

docs/scripts.rst
profxj: fixed

linetools/guis/utils.py
profxj: So added

nhmc:
Also, it's fine if you don't want to add documentation in this PR, but in that case could you open an issue about it so we remember to do it later?

nhmc:
Oh wait, I see you do have some docs - don't worry about opening the issue.

linetools/scripts/lt_continuumfit.py
profxj: Agreed

linetools/scripts/lt_continuumfit.py
profxj: Done

linetools/scripts/lt_continuumfit.py
profxj: Ah. I had been using my own convention
 -blah=
to set a value and
 --blah
for a flag.  

I kind of like my way but I suppose it isn't standard..

linetools/scripts/lt_xspec.py
profxj: Moved to redshift

profxj:
Addressed Neil's comments..

linetools/scripts/lt_continuumfit.py
nhmc: If you want it to be a flag (and not allow an argument), you can use the
action='store_true' keyword, see:
https://docs.python.org/2/library/argparse.html#action

profxj:
merging.

PR #184, opened 2016-02-02
Add 'y' navigation key, update docs

nhmc:
Adds a 'y' key to the navigation gui, which guesses the flux limits based on the visible wavelength range. Also updates the docs.

P.S. @profxj, in future when you update a notebook-generated versions of the rst doc files (e.g. xspecgui.rst), please update the original notebook instead, and then re-generate the rst file using nb_to_rst.src. Otherwise the notebook doesn't get updated.


profxj:
Looks good and got it.  I hadn't realized
that rst file was nb generated! :)

Merging.

PR #186, opened 2016-02-02
XSpectrum1D: Read DESI bricks

profxj:
and a bit more
docs
no tests

### DIFF
--- a/docs/xspectrum1d.rst
+++ b/docs/xspectrum1d.rst
@@ -111,3 +111,20 @@ pixel indices corresponding to a wavelength or velocity range, and
 the spectrum. For a complete list of all the available methods, see
 the API: `~linetools.spectra.xspectrum1d.XSpectrum1D`.
   
+Files Read
+==========
+
+Below is an table of the types of spectra files that
+can be read by `~linetools.spectra.io.readspec`.  If
+your file cannot be read, contact the authors and
+stand by.
+
+========================================================== =================
+Description                                                Instruments
+========================================================== =================
+simple 1D FITS files                                       ESI, HIRES, etc.
+binary FITS table from LowRedux                            LRIS,Kast,etc.
+multi-extension 1D FITS files from LowRedux                LRIS,Kast,etc.
+multi-extension binary FITS tables from PYPIT              LRIS,Kast,etc.
+brick files (2D images: flux, ivar; 1D image: wavelength)  DESI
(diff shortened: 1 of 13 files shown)


docs/xspectrum1d.rst
```diff
@@ -111,3 +111,20 @@ pixel indices corresponding to a wavelength or velocity range, and
 the spectrum. For a complete list of all the available methods, see
 the API: `~linetools.spectra.xspectrum1d.XSpectrum1D`.
   
+Files Read
```
nhmc: Maybe change to 'File Formats Read'.

linetools/spectra/io.py
```diff
@@ -236,12 +236,24 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, flux_tags=None,
                 tmpsig[gdp] = np.sqrt(1./sig[gdp])
                 sig = tmpsig
                 wave = 10.**wave
-    elif (head0['NAXIS'] == 2) and (head0['NAXIS2'] == 5): # SDSS .fit format
-        fx = hdulist[0].data[0,:].flatten()
-        sig = hdulist[0].data[2,:].flatten()
-        wave = setwave(head0)
+    elif head0['NAXIS'] == 2:
```
nhmc: Is there a FITS header keyword that can be used to show this is a DESI brick, instead of the file structure? Using a keyword seems less likely to produce unexpected bugs with unknown file types.

nhmc:
I left some inline comments - apart from those, it looks great. Merge when ready.

docs/xspectrum1d.rst
profxj: Done

linetools/spectra/io.py
profxj: Good suggestion.  I fear the current header is overly sparse(see below).
When they finally take real data that will change and I'll try to remember
to make this more unique.


SIMPLE  =                    T / conforms to FITS standard                      BITPIX  =                  -32 / array data type                                NAXIS   =                    2 / number of array dimensions                     NAXIS1  =                 2408                                                  NAXIS2  =                  100                                                  EXTEND  =                    T                                                  BRICKNAM= 'elg-100 '                                                            CHANNEL = 'z       '                                                            EXTNAME = 'FLUX    '           / no dimension                                   END

profxj:
Merging.

PR #187, opened 2016-02-03
XSpectrum1D: Minor correction to docstring.

ntejos:
Eventually, it would be nice to have the splice smart enough to realize which one is on the shorter/longer wavelength side...


nhmc:
I agree. Thanks for the doc edit (I think the incorrect description was my fault...).

PR #188, opened 2016-02-04
Refactor readspec

nhmc:
I found spectra.io.readspec very confusing. This refactors it so it's more clear what is going on (to me anyway, and hopefully other people who aren't @profxj...). 

Note that I also removed the ability to specify flux and wavelength tags, because this makes the code simpler, and because the list of tags in the docs was getting out of sync with the list used in the code.  It now just prints the list of tags if one of them can't be found.

It also makes ExamineSpecWidget plot the continuum by default.

### DIFF
--- a/docs/xspectrum1d.rst
+++ b/docs/xspectrum1d.rst
@@ -114,10 +114,10 @@ the API: `~linetools.spectra.xspectrum1d.XSpectrum1D`.
 File Formats Read
 =================
 
-Below is a table of the types of spectra files that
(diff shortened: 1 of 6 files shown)


profxj:
Looks great.  2 things to consider:

1.  Any document changes to make?

2.  Should we expand the tests, e.g. by adding in a few more test files?
This might be a good time as we are working on this code.

nhmc:
I added some docs. I will open an issue about testing more file formats.

PR #191, opened 2016-02-07
Refactor XSpectrum1D

profxj:
Lots and lots
Still needs docs
and a bug in FITS I/O makes read/write problematic right now


profxj:
All tests passing except numpy v1.9 and I have no idea why..

Neil, please review now.

linetools/spectra/xspectrum1d.py
```diff
@@ -27,12 +26,14 @@
 from ..analysis.interp import AkimaSpline
 eps = np.finfo(float).eps
 
+"""
 try:
     from specutils import Spectrum1D
```
ntejos: Get rid of this `try`.

linetools/spectra/xspectrum1d.py
```diff
@@ -40,22 +41,14 @@ class Spectrum1D(object):
 # Child Class of specutils/Spectrum1D
 # Generated by JXP to add functionality before it gets ingested in the
 # specutils distribution
```
ntejos: Get rid of these comments regarding specutils

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
     def from_tuple(cls, ituple, sort=True, **kwargs):
-        """Make an XSpectrum1D from a tuple of arrays.
+        """Make an XSpectrum1D from a tuple of numpy arrays or Quantity arrays
 
         Parameters
         ----------
         ituple : (wave,flux), (wave,flux,sig) or (wave,flux,sig,cont)
+          ndarray or Quantity array
           If wave is unitless, Angstroms are assumed
+          If flux is unitless, it is made dimensionless
```
ntejos: Mention that the units (if any) of sig and cont are assumed to be those of flux (?)

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
+        for ii in range(len(ltuple),4):
+            ltuple.append(None)
         if sort:
-            srt = np.argsort(uwave)
-            uwave = uwave[srt]
-            for i in range(1, len(ituple)):
-                if ltuple[i] is not None:
-                    ltuple[i] = ituple[i][srt]
+            srt = np.argsort(iwave)
+            iwave = iwave[srt]
```
ntejos: We should probably check that `iwave` and `iflux` have the same dimensions (same for sig and cont).

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
-            spec.co = None
-        elif len(ltuple) == 3:
-            spec = cls.from_array(uwave, u.Quantity(ltuple[1]),
-                uncertainty=StdDevUncertainty(ltuple[2]), **kwargs)
-            spec.co = None
-        else:
-            spec = cls.from_array(uwave, u.Quantity(ltuple[1]),
-                uncertainty=StdDevUncertainty(ltuple[2]), **kwargs)
-            spec.co = ltuple[3]
+        spec = cls(iwave, iflux, sig=ltuple[2], co=ltuple[3], units=dict(wave=wv_unit, flux=fx_unit), **kwargs)
```
ntejos: What is the default for the units of `sig` and `co` (?)

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
+        if sig is not None:
+            self._data['sig'] = np.reshape(sig, (self.nspec, self.npix))
+            self.flag_sig = True
+        else:
+            self.flag_sig = False
+        if co is not None:
+            self._data['co'] = np.reshape(co, (self.nspec, self.npix))
+            self.flag_co = True
+        else:
+            self._data['co'] = 1.
```
ntejos: Wouldn't it be better to set `self._data['co'] = None` if continuum is not defined?

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
-                spec.co = np.array(spec.co, copy=True)
+        """ Copy the spectrum
+        """
+        # Key components
+        data = self._data.copy()
+        units = self.units.copy()
+        meta = self.meta.copy()
+        #
+        new = XSpectrum1D(data['wave'], data['flux'], data['sig'], data['co'],
+                          units=units, meta=meta)
```
ntejos: Set `select=0` by default in the copy (?)

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
+        return new
 
-        return spec
+    @property
+    def wavelength(self):
+        """ Return the wavelength array with units
+        """
+        return self._data[self.select]['wave'] * self.units['wave']
```
ntejos: Nice!

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
+        """ Return the flux with units
+        """
+        flux =  self._data[self.select]['flux'] * self.units['flux']
+        if self.normed and self.flag_co:
+            flux /= self._data[self.select]['co']
+        return flux
+
+    @property
+    def co(self):
+        """ Return the flux with units
```
ntejos: flux -> continuum

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
+        """
+        if self.flag_co is False:
+            raise ValueError("This spectrum does not contain an input continuum array")
```
ntejos: `return None` (?)

linetools/spectra/xspectrum1d.py
```diff
@@ -241,28 +328,28 @@ def add_noise(self, seed=None, s2n=None):
...
 
         # Modify the flux
         if s2n is not None:
             sig = 1. / s2n
         else:
-            sig = self.sig
+            sig = self.sig.value
         #
-        self.flux = self.flux + (rand * sig) * self.flux.unit
+        self._data[self.select]['flux'] = self.flux.value + (rand * sig)
```
ntejos: Should we store the 'original' spectrum (i.e. flux values with no extra noise) somewhere (maybe something like `self._data['flux_orig']`)? I do not like the functionality of losing information without a way to recover the original one.

linetools/spectra/xspectrum1d.py
```diff
@@ -241,28 +328,28 @@ def add_noise(self, seed=None, s2n=None):
...
         """Set the uncertainty array to a constant value.
 
         Parameters
         ----------
-        sigv : float [0]
+        sigv : float, optional
           Scalar sigma value to use.
         """
-        self.uncertainty = StdDevUncertainty(np.ones(self.flux.size) * sigv)
+        self._data['sig'] = sigv
```
ntejos: Please impose that `sigv` has the same shape than the wavelength array.

linetools/spectra/xspectrum1d.py
```diff
@@ -276,46 +363,25 @@ def normalize(self, co=None, verbose=False, no_check=False):
...
-            if no_check:
-                print('WARNING: Continuum length differs from flux')
-                if len(co) > len(self.flux):
-                    self.flux /= co[0:len(self.flux)]
-                    if self.uncertainty is not None:
-                        self.uncertainty.array /= co[0:len(self.flux)]
-                    return
+        if co is not None:
+            if (len(co) != self.npix):
+                if no_check:
```
ntejos: Not convinced about the this `no_check` thing. Could you point me to an example where this `no_check` is useful ?

linetools/spectra/xspectrum1d.py
```diff
@@ -468,6 +497,10 @@ def plot(self, **kwargs):
 
         ax.set_xlim(xmin, xmax)
 
+        # Labels
+        ax.set_xlabel('Wavelength', size=16)
```
ntejos: Please add units in the labels.

linetools/spectra/xspectrum1d.py
```diff
@@ -578,16 +611,16 @@ def rebin(self, new_wv, do_sig=False):
...
                                    (new_dwv[1:]+dwv[ibad])/2)[0]
                 new_sig[bad_new] = 0.
-            # Init
-            newspec = XSpectrum1D.from_tuple((new_wv.value, new_fx, new_sig),
-                                             meta=self.meta.copy())
-            return newspec
         else:
-            return XSpectrum1D.from_array(new_wv, new_fx, meta=self.meta.copy())
+            new_sig = None
+        # Return
```
ntejos: Should this function rebin the continuum too, if it exists?

linetools/spectra/xspectrum1d.py
```diff
@@ -639,13 +672,15 @@ def box_smooth(self, nbox, preserve=False):
...
+                    self.sig[orig_pix], new_npix) / np.sqrt(nbox)
+            else:
+                new_sig = None
 
         # Return
-        return XSpectrum1D.from_array(
-            new_wv, new_fx, meta=self.meta.copy(),
-            uncertainty=apy.nddata.StdDevUncertainty(new_sig))
+        return XSpectrum1D.from_tuple(
+            (new_wv, new_fx, new_sig), meta=self.meta.copy())
```
ntejos: What about the continuum ?

ntejos:
The new XSpectrum1D looks great! I left some comments in for consideration before merging.

linetools/spectra/xspectrum1d.py
ntejos: Maybe return a new XSpectrum1D object instead? (similar to rebinning).

linetools/spectra/io.py
```diff
@@ -165,26 +163,37 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
...
+    if 'xspec1d' not in locals():
+        # Give Ang as default
+        if not hasattr(wave, 'unit'):
+            uwave = u.Quantity(wave, unit=u.AA)
+        elif wave.unit is None:
+            uwave = u.Quantity(wave, unit=u.AA)
+        else:
+            uwave = u.Quantity(wave)
+        xspec1d = XSpectrum1D.from_tuple((wave, fx, sig, None))
+
```
nhmc: Why is this if block needed? When I refactored io.py I made sure all the read blocks returned an XSpectrum1D. As long as any new reading code does the same, this check shouldn't be necessary.  I also think the code is clearer without it - to me it's simpler to require that all the read sections/functions return an xspec1d, rather than have wave fx and sig variables floating around.

linetools/spectra/io.py
```diff
@@ -165,26 +163,37 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
...
-            print('linetools.spectra.io.readspec(): Found continuum')
+        if isinstance(specfil,basestring):
+            if specfil.endswith('.fits'):
+                try:
+                    co = fits.getdata(specfil.replace('.fits', '_c.fits'))
+                except IOError:
+                    pass
+                else:
+                    xspec1d._data['co'] = co
+
```
nhmc: Is there any reason why you're reverting to the old co file code? I changed it to avoid a try-except statement, which is slow when the exception triggers (which it usually does in this case).

linetools/spectra/io.py
```diff
@@ -458,11 +476,15 @@ def parse_FITS_binary_table(hdulist, exten=None):
         return
     co_tags = ['CONT', 'CO', 'CONTINUUM', 'co', 'cont']
```
nhmc: Maybe add a co_tags keyword to match sig_tags and flux_tags.

linetools/spectra/io.py
```diff
@@ -458,11 +476,15 @@ def parse_FITS_binary_table(hdulist, exten=None):
...
+    # Finish
     xspec1d = XSpectrum1D.from_tuple((give_wv_units(wave), fx, sig, co))
+
+    if 'METADATA' in hdulist[0].header:
+        xspec1d.meta.update(json.loads(hdulist[0].header['METADATA']))
     return xspec1d
 
 def parse_linetools_spectrum_format(hdulist):
-    """ Parse a linetools-format spectrum from an hdulist
+    """ Parse an old linetools-format spectrum from an hdulist
```
nhmc: Why is this now an old format?  This format is fine for 2d flux/sig/wavelength arrays, we can simply save 2d arrays in each extension instead of 1d arrays.

linetools/spectra/io.py
```diff
@@ -473,60 +495,60 @@ def parse_linetools_spectrum_format(hdulist):
...
-        exten = 0
-    fx = hdulist[0].data[exten, :]
+    fx = hdulist[0].data
     # Sig
-    ivar = hdulist[1].data[exten, :]
-    sig = np.zeros(ivar.size)
+    ivar = hdulist[1].data
+    sig = np.zeros_like(ivar)
     gdi = ivar > 0.
     sig[gdi] = np.sqrt(1./ivar[gdi])
```
nhmc: I suggest setting bad error values to np.nan. This seems to show more clearly that the error is not meaningful (and it also means those points aren't plotted by matplotlib).

linetools/spectra/io.py
```diff
@@ -473,60 +495,60 @@ def parse_linetools_spectrum_format(hdulist):
...
     # Wave
     wave = hdulist[2].data
     wave = give_wv_units(wave)
-    xspec1d = XSpectrum1D.from_tuple((wave, fx, sig, None))
+    wave = np.outer(np.ones(fx.shape[0]), wave)
```
nhmc: Maybe replace this with `np.tile(wave, (fx.shape[0],1))`  which looks clearer to me (and is a bit faster).

linetools/spectra/xspectrum1d.py
```diff
@@ -40,22 +41,14 @@ class Spectrum1D(object):
...
-        transformation between pixel coordinates and "dispersion" coordinates
-        this carries the unit of the dispersion
-
-    unit : `~astropy.unit.Unit` or None, optional
-        unit of the flux, default=None
+    _data : `~numpy.ndarray`
+        Structured array containing all of the data
+        This can be a set of 1D spectra
 
     mask : `~numpy.ndarray`, optional
```
nhmc: Does the code still support masks (I don't think it does)? If not remove this comment.

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
           Filename
         """
         # put import here to avoid circular import with io.py
-        from .io import readspec
-        slf = readspec(ifile, **kwargs)
+        #from .io import readspec
+        from linetools.spectra import io as tio
+        reload(tio)
```
nhmc: remove this (or use imp.reload for python 3 compatibility)

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
-                  uncertainty=spec1d.uncertainty, mask=spec1d.mask,
-                  meta=spec1d.meta.copy())
-        slf.co = (spec1d.co if hasattr(spec1d, 'co') else None)
 
+        Avoids error array for now
+        """
+        slf = cls.from_tuple((spec1d.dispersion, spec1d.flux))
+        #slf = cls(flux=spec1d.flux, wcs=spec1d.wcs, unit=spec1d.unit,
+        #          uncertainty=spec1d.uncertainty, mask=spec1d.mask,
+        #          meta=spec1d.meta.copy())
```
nhmc: I think the `from_spec1d` method can be removed altogether, we can use `copy` instead. (Right?)

linetools/spectra/xspectrum1d.py
```diff
@@ -84,92 +75,205 @@ def from_file(self, ifile, **kwargs):
...
+                               dtype=[(str('wave'), 'float64', (self.npix)),
+                                      (str('flux'), 'float32', (self.npix)),
+                                      (str('sig'),  'float32', (self.npix)),
+                                      (str('co'),   'float32', (self.npix)),
+                                     ])
+        self._data['wave'] = np.reshape(wave, (self.nspec, self.npix))
+        self._data['flux'] = np.reshape(flux, (self.nspec, self.npix))
+        if sig is not None:
+            self._data['sig'] = np.reshape(sig, (self.nspec, self.npix))
+            self.flag_sig = True
```
nhmc: Why is a flag needed? Can't you just test with `if 'sig' in data`? (similarly for co)

linetools/spectra/xspectrum1d.py
nhmc: I agree.

linetools/spectra/xspectrum1d.py
nhmc: (or if you always want sig/co keywords to be present in data, then you could test with `if data['sig'] is not None`)

linetools/spectra/xspectrum1d.py
```diff
@@ -702,39 +736,46 @@ def splice(self, spec2, wvmx=None, scale=1.):
...
         table.write(outfil, format=format)
 
     # Write to fits
-    def write_to_fits(self, outfil, clobber=True, add_wave=False):
-        ''' Write to a FITS file.
+    def write_to_fits(self, outfil, clobber=True):
+        """ Write to a FITS file.
 
-        Note that this does not generate a binary FITS table format.
+        This generates a binary FITS table format.
```
nhmc: Why make this a binary fits table? Can't we just stick with the old format but use 2d arrays for everything instead of 1d arrays?

linetools/spectra/xspectrum1d.py
```diff
@@ -742,22 +783,24 @@ def write_to_fits(self, outfil, clobber=True, add_wave=False):
...
+        #
+        tblhdu = fits.BinTableHDU(self._data, name='DATA')
+        # List
+
+        hdulist = fits.HDUList([prihdu, tblhdu])
 
+        """
         # Type
-        if type(self.wcs) is Spectrum1DPolynomialWCS:  # CRVAL1, etc. WCS
+        if type(self.wcs) is False:#Spectrum1DPolynomialWCS:  # CRVAL1, etc. WCS
```
nhmc: We can just pull out this WCS stuff altogether right? i.e. we will only support writing an array of wavelengths.

linetools/spectra/xspectrum1d.py
```diff
@@ -815,11 +855,18 @@ def write_to_fits(self, outfil, clobber=True, add_wave=False):
...
+        # META
         if self.meta is not None and len(self.meta) > 0:
             d = liu.jsonify(self.meta)
             prihdu.header['METADATA'] = json.dumps(d)
 
-        hdu.writeto(outfil, clobber=clobber)
+        # Units
+        units = self.units.copy()
+        d = liu.jsonify(units)
+        prihdu.header['UNITS'] = json.dumps(d)
```
nhmc: Neat! If a bit hacky :)

linetools/spectra/xspectrum1d.py
```diff
@@ -40,22 +41,14 @@ class Spectrum1D(object):
...
-
-    wcs : `spectrum1d.wcs.specwcs.BaseSpectrum1DWCS`-subclass
-        transformation between pixel coordinates and "dispersion" coordinates
-        this carries the unit of the dispersion
-
-    unit : `~astropy.unit.Unit` or None, optional
-        unit of the flux, default=None
+    _data : `~numpy.ndarray`
+        Structured array containing all of the data
+        This can be a set of 1D spectra
```
nhmc: Maybe we can remove the leading underscore from `_data`? It think it's ok if users see and access this directly.

nhmc:
I left some inline comments. Note that I strongly prefer not changing to a binary FITS table format.

Not sure what the numpy 1.9 test failure is. Do you need to fix it now? If not, I'll hold off investigating until you've made changes addressing our comments.

linetools/spectra/xspectrum1d.py
profxj: done.

linetools/spectra/xspectrum1d.py
profxj: done

linetools/spectra/xspectrum1d.py
profxj: Noted.

linetools/spectra/xspectrum1d.py
profxj: This is checked in the _init.  Have added
tests for sig and co there now.

linetools/spectra/xspectrum1d.py
profxj: These are the same units as flux.  I've noted this now.

linetools/spectra/xspectrum1d.py
profxj: If I use None, numpy turns these into Nan's in the data array,
which makes me a bit queasy..  
I've now just not touched the values (no longer setting to 1. either).

linetools/spectra/xspectrum1d.py
profxj: select=self.select 

good catch

linetools/spectra/xspectrum1d.py
profxj: edited

linetools/spectra/xspectrum1d.py
profxj: Now only throws a warning and returns None.

linetools/spectra/xspectrum1d.py
profxj: Yes, I agree.  Done.

linetools/spectra/xspectrum1d.py
profxj: sigv is a scalar.

linetools/spectra/xspectrum1d.py
profxj: I have some innocuous SDSS analyzed data for QPQ.  I may
be able to get rid of this before too long but need it for now..

linetools/spectra/xspectrum1d.py
profxj: I've done so for wavelength.  Am avoiding it for Flux for
now as it is a bit of a pain when the unit is dimensionless.

linetools/spectra/xspectrum1d.py
profxj: I leave that to future development.

linetools/spectra/xspectrum1d.py
profxj: Not included in this method (thus far)

linetools/spectra/io.py
profxj: I believe it was because of a bad merge (I fear I didn't pull your refactor
until I started fiddling about).

Am commenting this block out now and if things are ok will kill it altogether.

linetools/spectra/io.py
profxj: added a co_tag keyword

linetools/spectra/io.py
profxj: Another merge error..
I've reverted back to yours.

linetools/spectra/io.py
profxj: Ok, let me give more thought to the writing..

linetools/spectra/io.py
profxj: I wasn't familiar with tile.  outer was an IDL approach! ;)

linetools/spectra/xspectrum1d.py
profxj: Removed

linetools/spectra/xspectrum1d.py
profxj: removed.

linetools/spectra/io.py
profxj: I worry about writing np.nan to a FITS file
(or any other format for that matter).   I could
set to -1.

linetools/spectra/xspectrum1d.py
profxj: this is if someone is using specutils

linetools/spectra/xspectrum1d.py
profxj: converted to data from _data

linetools/spectra/xspectrum1d.py
profxj: Removed

linetools/spectra/xspectrum1d.py
profxj: I now have 2 write FITS procedures.  One for binary FITS table.

linetools/spectra/xspectrum1d.py
profxj: I've now gone with np.nan and isnan.

nhmc:
Thanks, this looks good. 

Thanks for restoring the old FITS writing method. But what is the motivation for having a binary FITS format at all? It's simpler to have just a single FITS spectrum format, and the non-binary table version is more flexible.

nhmc:
Also, can you update the installation docs to remove mentions of specutils? (otherwise I can do it after you merge).

linetools/spectra/xspectrum1d.py
```diff
@@ -219,103 +305,82 @@ def set_diagnostics(self):
...
 
         # Modify the flux
         if s2n is not None:
             sig = 1. / s2n
         else:
-            sig = self.sig
+            sig = self.sig.value
+        # Copy
+        newspec = self.copy()
+        newspec.data[self.select]['flux'] = self.flux.value + (rand * sig)
```
nhmc: This should be 

```
newspec.data['flux'][self.select] = self.flux.value + (rand * sig)
```

Fixing this should make the numpy 1.9 test pass.

linetools/spectra/xspectrum1d.py
nhmc: Ok, could you add to the docstring that it should take a specutils.Spectrum1D as input?

profxj:
I prefer to leave the FITS Table writing method.   The data are
in a table already so that makes a bit more sense to me.

Have removed specutils from the install.

Making that bug fix now..

linetools/spectra/xspectrum1d.py
profxj: I'd be surprised if that worked, but am trying..

profxj:
Merging (before tests fail again!)..  :)

PR #194, opened 2016-02-18
XSpectrum1D: make sure sigma is well defined for gaussian smoothing.

ntejos:
XSpectrum1D: make sure sigma is well defined for gaussian smoothing.


nhmc:
Merging. 

I think we should just set sig = None rather than having a sig_is_set flag, but that can be fixed later.

ntejos:
I agree with `sig = None` when it does not exist.

PR #197, opened 2016-02-22
Allow assigning values to XSpectrum1D attributes

nhmc:
This allows assigning arrays directly to the wavelength, flux, sig and co attributes. e.g.

```python
wa = np.arange(10)
fl = np.ones_like(wa)
sp = XSpectrum1d.from_tuple((wa, fl))
wave = np.arange(1, len(sp.wavelength) + 1)
sp.wavelength = wave
```

Units are updated for the wavelength or flux (if the assigned array is a Quantity), and ignored for sig and co.

Tests are included.


nhmc:
Please don't merge yet, I think there's better way to handle this. I will refactor when I get time.

profxj:
Nice! Would have never guessed you could do that. :)

PR #198, opened 2016-02-26
XSpectrum1D: solves bug (reported by Joe Ribaudo) with the continuum …

ntejos:
XSpectrum1D: solves bug (reported by Joe Ribaudo) with the continuum fitting when self.co is set to None.


profxj:
Looks good.

Merging..

PR #199, opened 2016-03-02
Scripts: avoid scientific notation when possible for axes tick labels.

ntejos:
I find scientific notation annoying in the wavelength axis...


nhmc:
Look good, thanks. Merging.

PR #200, opened 2016-03-07
AbsSystem/DLASystem: from_components

profxj:
Puts a 'hook' into AbsSystem specific to DLASystem
which comes from pyigm.

Doesn't generate a dependence of pyigm for linetools,
but is not an elegant approach.  Might be a first
step towards pushing AbsSystem into pyigm..

### DIFF
--- a/linetools/isgm/abssystem.py
+++ b/linetools/isgm/abssystem.py
@@ -106,7 +106,16 @@ class AbsSystem(object):
         init_comp = components[0]
         if vlim is None:
             vlim = init_comp.vlim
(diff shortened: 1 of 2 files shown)


nhmc:
Why make this DLA-specific?  Instead of checking the class name, it seems better to just check whether any component has Zion = (1,1).

profxj:
Have made the NHI call generic.

nhmc:
Looks good to me. Please merge when the tests pass.

nhmc:
Merging.

PR #201, opened 2016-03-09
LSF: COS life-time position 3 LSFs added.

ntejos:
LSF: COS life-time position 3 LSFs added.


nhmc:
This looks good to me. Please merge after tests pass.

I restarted the test suite, it seemed to have failed by timing out.

ntejos:
The new COS LP3 tabulated LSFs, have actually 1 column per Angstroms, which means a lot of columns! So my original approach based on interpolating over all tabulated columns was not fast enough anymore. Now I am finding the 3 closest columns to a given wavelength and interpolating using only these, which should be more efficient. The code is not the prettiest but does the job; although if someone suggest how to make the code cleaner I would be very happy to incorporate it before merging.

PR #203, opened 2016-03-25
xspectrum1d: rebin: rebins continuum if `co_is_set`.

ntejos:
xspectrum1d: rebin: re-interpolate continuum if `co_is_set` to the new wv array.

### DIFF
--- a/linetools/spectra/xspectrum1d.py
+++ b/linetools/spectra/xspectrum1d.py
@@ -539,7 +539,7 @@ or QtAgg backends to enable all interactive plotting commands.
         npix = len(self.wavelength)
         wvh = (self.wavelength + np.roll(self.wavelength, -1)) / 2.
         wvh[npix - 1] = self.wavelength[npix - 1] + \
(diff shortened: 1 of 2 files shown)


linetools/spectra/xspectrum1d.py
```diff
@@ -1030,7 +1039,7 @@ def fit_continuum(self, knots=None, edges=None, wlim=None, dw=10.,
             [tuple(pts) for pts in wrapper.contpoints])
         self.meta['contpoints'].sort()
 
-    def _interp_continuum(self, x, y):
+    def _interp_continuum(self, x, y, wv):
```
nhmc: Can you make `wv` a keyword argument with default value `None`, meaning that the spectrum wavelength is used? It seems fine to have a wavelength argument, but I think it makes sense to use the spectrum wavelength array of the spectrum by default.

nhmc:
Apart from the inline comment, this looks great to me.

nhmc:
Thanks. Merging.

PR #205, opened 2016-03-30
AbsLine Kinematics

profxj:
Code for measuring AbsLine kinematics

A number of bits and pieces were added along the way
to deal with AbsSystems

Tests and (minimal) docs included


linetools/analysis/abskin.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    # Velocity array
+    npix = len(velo)
+    pix = np.arange(npix).astype(int)
+
+    # Calculate dv
+    dv = np.abs(np.median(velo-np.roll(velo,1)))
+
+    # Test for bad pixels
+    badzero = np.where((flux == 0) & (sig <= 0))[0]
```
nhmc: Should this be `(flux == 0) | (sig <= 0)`? Also maybe change to ~(sig > 0) so it will catch NaN values.

nhmc:
This looks fine to me. Please merge after considering the inline comment.

profxj:
Good catch!

Added a couple of tests just to increase coverage.

profxj:
merging

PR #206, opened 2016-04-04
Allow an astropy table column for wavelengths in from_tuple

nhmc:
Fixes a bug in `from_tuple` if the input wavelength array object has a `unit` attribute but no `value` attribute (which is possible for astropy table columns.)


nhmc:
Merging

PR #207, opened 2016-04-07
Generate normalized spectrum

profxj:
I am finding it useful on occasion to generate a distinct 
spectrum object that is the normalized version.

Added a test too.


profxj:
merging

PR #210, opened 2016-04-11
Tag keywords for lt_xspec

profxj:
This is a bit unwieldy, but I'm not sure if there is a better 
approach.  Please suggest one if you see it.

### DIFF
--- a/linetools/spectra/io.py
+++ b/linetools/spectra/io.py
@@ -413,7 +413,8 @@ def parse_UVES_popler(hdulist):
     return xspec1d
 
 def parse_FITS_binary_table(hdulist, exten=None, wave_tag=None, flux_tag=None,
-                            sig_tag=None, co_tag=None, var_tag=None):
+                            sig_tag=None, co_tag=None, var_tag=None,
+                            ivar_tag=None):
     """ Read a spectrum from a FITS binary table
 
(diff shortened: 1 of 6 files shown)


linetools/spectra/io.py
```diff
@@ -412,18 +412,20 @@ def parse_UVES_popler(hdulist):
     xspec1d = XSpectrum1D.from_tuple((uwave, fx, sig, co))
     return xspec1d
 
-def parse_FITS_binary_table(hdulist, exten=None, flux_tag=None,
-                            sig_tag=None, co_tag=None):
+def parse_FITS_binary_table(hdulist, exten=None, wave_tag=None, flux_tag=None,
+                            sig_tag=None, co_tag=None, var_tag=None):
```
nhmc: Does an ivar_tag also need to be added here? It's listed above as an option in `lt_xspec`.

nhmc:
Thanks. I left an inline comment.
 
It might simplify things to take the spectrum file-reading code outside ExamineSpecWidget, and make ExamineSpecWidget require an XSpectrum1D as input. But I'm not sure - only explore that if you feel like it.

Please merge when you feel it's ready.

profxj:
Merging

PR #212, opened 2016-05-05
ExamineSpec

profxj:
edits to ExamineSpec to deal with continuum changes

No new tests or docs, but used on real spectra


nhmc:
I will trust that you know what you're doing here :)  Merging.

PR #214, opened 2016-05-10
Synth_comp

profxj:
Some small edits for synthesizing components with lines that
don't contribute to the column.

And a test modification

### DIFF
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -397,7 +397,7 @@ class AbsComponent(object):
         self.flag_N = 0
         for aline in self._abslines:
             if aline.attrib['flag_N'] == 0:  # No value
-                warnings.warn("Absline {} has flag=0.  Hopefully you expected that")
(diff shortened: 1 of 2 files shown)


linetools/isgm/abscomponent.py
```diff
@@ -396,6 +396,10 @@ def synthesize_colm(self, overwrite=False, redo_aodm=False, **kwargs):
         # Collate
         self.flag_N = 0
         for aline in self._abslines:
+            if aline.attrib['flag_N'] == 0:  # No value
+                warnings.warn("Absline {} has flag=0.  Hopefully you expected that")
```
nhmc: Did you mean to print the absorption line info here? There is a '{}' but no argument to print.

nhmc:
Apart from the inline comment, this looks great. Please merge when ready.

linetools/isgm/abscomponent.py
profxj: Indeed!

profxj:
Merging.

PR #216, opened 2016-05-19
Zero sig

profxj:
Small edit to avoid crash in readspec

Also a few minor edits to linelist


nhmc:
This looks fine to me. Merging.

PR #217, opened 2016-06-01
Remove a specutils import

nhmc:
Also allow passing arbitrary keyword arguments to XSpectrum1D.box_smooth(), to be passed on to astropy.convolution.convolve.


nhmc:
It seems to be failing tests on the numpy 1.9 build due a scipy import fail. I'll take a look at the reason for this tomorrow.

profxj:
These look good to me.

Merge if you are fine with the numpy 1.9 issue (I am).

nhmc:
Ok, I will punt on the numpy 1.9 issue for now.

PR #218, opened 2016-06-02
Odds and Ends: AbsSys, Python3 

profxj:
Series of minor edits to:
  a.  Refactor abs_type in AbsSystem
  b.  Deal with some syntax/bugs when running in Python 3
  c.  Some GUI edits
  d.  Added a test to cover AbsSystem.from_dict and from_json
  e.  Addresses 'crash' when running measure_EW or measure_AODM on a 'bad' spectrum

### DIFF
--- a/linetools/isgm/abssystem.py
+++ b/linetools/isgm/abssystem.py
@@ -371,6 +371,26 @@ class AbsSystem(object):
         # Return
         return outdict
 
+    def write_json(self, outfil=None):
+        """ Generate a JSON file from the system
+
+        Returns
+        -------
+
+        """
+        import io, json
+        # Generate the dict
+        odict = self.to_dict()
+        # Write
+        if outfil is None:
+            outfil = self.name+'.json'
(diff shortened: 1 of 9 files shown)


linetools/isgm/abssystem.py
```diff
@@ -91,7 +91,7 @@ def from_abslines(cls, abslines, vlim=None, **kwargs):
         return slf
 
     @classmethod
-    def from_components(cls, components, vlim=None):
+    def from_components(cls, components, vlim=None, NHI=None):
         """Instantiate from a list of AbsComponent objects
 
         Parameters
```
ntejos: Please add a note about `NHI` parameter in docstring.

ntejos:
Left a minor comment in; all the rest look fine to me.

linetools/isgm/abssystem.py
profxj: Added

profxj:
Merging..

PR #219, opened 2016-06-02
XAbsSysGui

profxj:
Here is a new GUI that I would tend to use to 
examine and modify the lines of an AbsSystem.
It has some similar concepts to IGMGuesses but
is only 1 system at a time.

Have tested it on a DLA but there are no formal tests
with this PR.  

Have included some documentation and can be pushed to
include more.

### DIFF
--- a/docs/guis.rst
+++ b/docs/guis.rst
@@ -14,6 +14,7 @@ Notebooks
 
    Continuum fitting <Interactive_continuum_fitting_example>
    XSpecGui <xspecgui>
+   XAbsSysGui <xabssysgui>
 
 Overview
 ========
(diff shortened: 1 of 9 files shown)


docs/guis.rst
```diff
@@ -42,4 +43,11 @@ This enables visual inspection of a spectrum.  Simple analysis
 notebook for details.
 
 
+XAbsSysGui
+==========
 
+This shows a velocity (stack) plot of the absorption lines from
+an input absorption line system.  The user can then modify the
+velocity limits that would be used for subsequent analysis, flag
+bad lines, blends, set limits, etc.  The modified absoprtion system
```
ntejos: absoprtion -> absorption

nhmc:
I'm happy you you to merge when @ntejos has had a look.

profxj:
Merging..

PR #220, opened 2016-06-03
SpecLine Items

profxj:
Enables one to grab items from a Specline by
overloading get_item, e.g.

absline['EW']

instead of 

absline.attrib['EW']

Just for convenience

Tests and docs too

### DIFF
--- a/linetools/spectralline.py
+++ b/linetools/spectralline.py
@@ -543,7 +543,23 @@ class AbsLine(SpectralLine):
         # Log
         laa.log_clm(self.attrib)
 
-    # Output
+    def __getitem__(self, k):
+        """  Passback a data bit from the Line
+        """
+        try:
+            return getattr(self,k)
+        except AttributeError:
+            try:
+                return self.attrib[k]
+            except KeyError:
+                try:
+                    return self.analy[k]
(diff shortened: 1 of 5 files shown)


linetools/spectralline.py
```diff
@@ -543,7 +543,23 @@ def measure_aodm(self, nsig=3., normalize=True):
...
+                return self.attrib[k]
+            except KeyError:
+                try:
+                    return self.analy[k]
+                except KeyError:
+                    try:
+                        return self.data[k]
+                    except KeyError:
+                        return None
+
```
nhmc: I worry that this much flexibility will produce bugs that are difficult to track down. Do you need to be able to access everything here (analy, attrib, & data)? Or would just having one or two deal with most use cases?

profxj:
I should have added that I feel this functionality should be used
very rarely when coding 'internally' with linetools.  It is primarily
meant as a convenience for the user to access data from a line
outside of linetools.  

I could live without the data dict and maybe the getattr piece if that were
a worthwhile compromise.

profxj:
I've decided to close this PR.  Am going to do this
differently (and in a safer way).

PR #222, opened 2016-06-04
Relative Abundances Class

profxj:
Class for handling relative abundances.  
Mainly for an AbsSystem.  And maybe 
mainly for DLAs

Tests and doc strings too

Will add a short Notebook and 
an .rst file before merging.

### DIFF
--- a/docs/relabund.rst
+++ b/docs/relabund.rst
@@ -1,0 +1,66 @@
+.. _RelAbund:
+
+**************
+RelAbund Class
+**************
+
+.. index:: RelAbund
+
+Notebooks
+=========
+
+.. toctree::
+   :maxdepth: 1
+
+   Simple Examples <RelAbund>
+
+Overview
+========
+
+This class packages the relative abundances of an object, typically
+an AbsSystem.
+
+Instantation
+============
+
+To begin, make a new class instance::
+
+   XY = RelAbund()
+   Loading abundances from Asplund2009
+
+Then load data into the data dict.  Here is an example::
+
+   XY._data = {6: dict(flag=1, XH=-1., sigXH=0.2, sig=0.05),
+            8: dict(flag=2, XH=-1.4, sigXH=0.25, sig=0.05),
+            14: dict(flag=1, XH=-1.1, sigXH=0.25, sig=0.05),
+            26: dict(flag=1, XH=-1.4, sigXH=0.25, sig=0.05),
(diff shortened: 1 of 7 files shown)


docs/relabund.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Then load data into the data dict.  Here is an example::
+
+   XY._data = {6: dict(flag=1, XH=-1., sigXH=0.2, sig=0.05),
+            8: dict(flag=2, XH=-1.4, sigXH=0.25, sig=0.05),
+            14: dict(flag=1, XH=-1.1, sigXH=0.25, sig=0.05),
+            26: dict(flag=1, XH=-1.4, sigXH=0.25, sig=0.05),
+            32: dict(flag=3, XH=-0.8, sigXH=0.25, sig=0.05),
+            }
+
```
nhmc: What do the flag values mean? They should be described here. Ideally the flag values could be short strings that suggest what they are. e.g. 'lo', 'hi' (or whatever) for lower and upper limits. (But I realise this means making changes elsewhere)

docs/relabund.rst
```diff
@@ -0,0 +1,66 @@
...
+More frequent usage will be to instantate using an input
+table of column density measurements, e.g.::
+
+   dla.XY = RelAbund.from_ionclm_table((1,dla.NHI, dla.sig_NHI[0]), dla._ionN)
+
+See pyigm DLA abund Notebook for more.
```
nhmc: Can you link to this here? I couldn't find it with a quick search.

nhmc:
Thanks for the PR. 

Can you give an example of how you would use this to do a quick calculation of an abundance given two column densities? In barak, for example, you do this via:

``` barak.abundances.calc_abund('C', 'Si', 13.5, 13.2)```

docs/relabund.rst
profxj: Have added a Table in the doc.

Also, the table() method adds a dict to its meta 
that describes the flagging.

docs/relabund.rst
profxj: It is still in a branch..  
I'll try to remember to do this when I merge it.

profxj:
Have just added a new instantiation method which also prints the value
of the pair inputted:

CSi = RelAbund.from_clm_pair('C', 13.5, 'Si', 13.2)

See the updated Notebook

nhmc:
Looks good, thanks for the changes. Please merge when you're ready.

profxj:
merging.

PR #223, opened 2016-06-09
Remove numpy 1.9 from travis testing

nhmc:
There's a problem with travis testing of numpy version 1.9, so this removes that test. I don't think we need to worry about supporting 1.9 anymore anyway, later stable versions are now in anaconda.


nhmc:
Merging.

PR #226, opened 2016-06-17
Xspec1d -- Masked arrays

profxj:
Refactor XSpectrum1D to use masked numpy arrays 

This is primarily for handling a set of spectra with variable
pixel number.  I don't think it will slow/harm performance for our standard
single spectrum usage.  

At present, you have to initiate masking by setting mask_edges=True, but I am
considering that as the default.

Give me your thoughts on this one..

### DIFF
--- a/linetools/spectra/xspectrum1d.py
+++ b/linetools/spectra/xspectrum1d.py
@@ -134,7 +134,7 @@ class XSpectrum1D(object):
         return spec
 
     def __init__(self, wave, flux, sig=None, co=None, units=None, select=0,
-                 meta=None, verbose=False):
+                 meta=None, verbose=False, mask_edges=False):
         """
         Parameters
         ----------
@@ -146,6 +146,8 @@ class XSpectrum1D(object):
           Required keys are 'wave' and 'flux'
         select : int, optional
           Selected Spectrum
+        mask_edges : bool, optional
+          Mask the edges of each spectrum where sig <= 0.
         """
         # Error checking
         if not isinstance(wave, np.ndarray):
@@ -169,7 +171,7 @@ class XSpectrum1D(object):
                 self.nspec, self.npix))
 
         # Data array
-        self.data = np.empty((self.nspec,), #self.npix),
+        self.data = np.ma.empty((self.nspec,), #self.npix),
                                dtype=[(str('wave'), 'float64', (self.npix)),
                                       (str('flux'), 'float32', (self.npix)),
                                       (str('sig'),  'float32', (self.npix)),
@@ -182,6 +184,12 @@ class XSpectrum1D(object):
             if wave.shape[0] != sig.shape[0]:
                 raise IOError("Shape of wave and sig vectors must be identical")
             self.data['sig'] = np.reshape(sig, (self.nspec, self.npix))
+            if mask_edges:
+                for kk in range(self.nspec):
+                    sigval = np.where(self.data['sig'][kk] > 0.)[0]
+                    for key in self.data.dtype.names:
+                        self.data[key][kk][0:sigval[0]].mask = True
+                        self.data[key][kk][sigval[-1]+1:].mask = True
         else:
             self.data['sig'] = np.nan
         if co is not None:
@@ -231,7 +239,8 @@ class XSpectrum1D(object):
     def wavelength(self):
         """ Return the wavelength array with units
         """
-        return self.data[self.select]['wave'] * self.units['wave']
+        #return self.data[self.select]['wave'] * self.units['wave']
+        return self.data['wave'][self.select].compressed() * self.units['wave']
 
     @wavelength.setter
     def wavelength(self, value):
@@ -243,9 +252,10 @@ class XSpectrum1D(object):
     def flux(self):
         """ Return the flux with units
         """
-        flux =  self.data[self.select]['flux'] * self.units['flux']
+        #flux =  self.data[self.select]['flux'] * self.units['flux']
+        flux =  self.data['flux'][self.select].compressed() * self.units['flux']
         if self.normed and self.co_is_set:
-            flux /= self.data[self.select]['co']
+            flux /= self.data['co'][self.select]
         return flux
 
     @flux.setter
@@ -258,7 +268,8 @@ class XSpectrum1D(object):
     def sig_is_set(self):
         """ Returns whether the error array is set
         """
-        if np.isnan(self.data[self.select]['sig'][0]):
+        #if np.isnan(self.data[self.select]['sig'][0]):
+        if np.isnan(self.data['sig'][self.select][0]):
             return False
         else:
             return True
@@ -271,9 +282,10 @@ class XSpectrum1D(object):
             warnings.warn("This spectrum does not contain an input error array")
             return np.nan
         #
-        sig = self.data[self.select]['sig'] * self.units['flux']
+        #sig = self.data[self.select]['sig'] * self.units['flux']
+        sig = self.data['sig'][self.select].compressed() * self.units['flux']
         if self.normed and self.co_is_set:
-            sig /= self.data[self.select]['co']
+            sig /= self.data['co'][self.select]
         return sig
 
     @sig.setter
@@ -287,7 +299,7 @@ class XSpectrum1D(object):
     def co_is_set(self):
         """ Returns whether a continuum is defined
         """
-        if np.isnan(self.data[self.select]['co'][0]):
+        if np.isnan(self.data['co'][self.select][0]):
             return False
         else:
             return True
@@ -299,7 +311,7 @@ class XSpectrum1D(object):
         if not self.co_is_set:
             warnings.warn("This spectrum does not contain an input continuum array")
             return np.nan
-        return self.data[self.select]['co'] * self.units['flux']
+        return self.data[self.select]['co'].compressed() * self.units['flux']
 
     @co.setter
     def co(self, value):
@@ -326,7 +338,6 @@ class XSpectrum1D(object):
             self.set_diagnostics()
             return self._wvmax
 
-
     def set_diagnostics(self):
         """Generate simple diagnostics on the spectrum.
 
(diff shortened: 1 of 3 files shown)


linetools/spectra/xspectrum1d.py
```diff
@@ -143,9 +143,11 @@ def __init__(self, wave, flux, sig=None, co=None, units=None, select=0,
         sig : ndarray, optional
         units : dict, optional
           Dict containing the units of wavelength, flux
-          Required keys are 'wave' and 'flux'
+          Required keys are 'wave' and 'flux li
         select : int, optional
           Selected Spectrum
+        mask_edges : bool, optional
+          Mask the edges of each spectrum where sig <= 0.
```
ntejos: Does this also include gaps in the spectrum?

linetools/spectra/xspectrum1d.py
```diff
@@ -143,9 +143,11 @@ def __init__(self, wave, flux, sig=None, co=None, units=None, select=0,
         sig : ndarray, optional
         units : dict, optional
           Dict containing the units of wavelength, flux
-          Required keys are 'wave' and 'flux'
+          Required keys are 'wave' and 'flux li
```
ntejos: flux li?

linetools/spectra/xspectrum1d.py
```diff
@@ -182,6 +184,12 @@ def __init__(self, wave, flux, sig=None, co=None, units=None, select=0,
             if wave.shape[0] != sig.shape[0]:
                 raise IOError("Shape of wave and sig vectors must be identical")
             self.data['sig'] = np.reshape(sig, (self.nspec, self.npix))
+            if mask_edges:
+                for kk in range(self.nspec):
+                    sigval = np.where(self.data['sig'][kk] > 0.)[0]
+                    for key in self.data.dtype.names:
+                        self.data[key][kk][0:sigval[0]].mask = True
+                        self.data[key][kk][sigval[-1]+1:].mask = True
```
ntejos: This will only work for masking out edges, I think.

How about define a boolean array: `valid_sigma = self.data['sig'][kk] > 0.`, and then use the negation of this as the mask? e.g. `self.data[key][kk][0:sigval[0]].mask = ~valid_sigma`

linetools/spectra/xspectrum1d.py
```diff
@@ -231,7 +239,8 @@ def copy(self):
     def wavelength(self):
         """ Return the wavelength array with units
         """
-        return self.data[self.select]['wave'] * self.units['wave']
+        #return self.data[self.select]['wave'] * self.units['wave']
+        return self.data['wave'][self.select].compressed() * self.units['wave']
```
ntejos: Not sure about compressed. If the spectrum has gaps, not sure how this compressed data would look like.

linetools/spectra/xspectrum1d.py
```diff
@@ -243,9 +252,10 @@ def wavelength(self, value):
     def flux(self):
         """ Return the flux with units
         """
-        flux =  self.data[self.select]['flux'] * self.units['flux']
+        #flux =  self.data[self.select]['flux'] * self.units['flux']
+        flux =  self.data['flux'][self.select].compressed() * self.units['flux']
```
ntejos: Maybe in this case instead of the compressed data, retrieve the masked values as zeros instead?

linetools/spectra/xspectrum1d.py
```diff
@@ -271,9 +282,10 @@ def sig(self):
             warnings.warn("This spectrum does not contain an input error array")
             return np.nan
         #
-        sig = self.data[self.select]['sig'] * self.units['flux']
+        #sig = self.data[self.select]['sig'] * self.units['flux']
+        sig = self.data['sig'][self.select].compressed() * self.units['flux']
```
ntejos: Ditto as in for flux (above).

linetools/spectra/xspectrum1d.py
```diff
@@ -299,7 +311,7 @@ def co(self):
         if not self.co_is_set:
             warnings.warn("This spectrum does not contain an input continuum array")
             return np.nan
-        return self.data[self.select]['co'] * self.units['flux']
+        return self.data['co'][self.select].compressed() * self.units['flux']
```
ntejos: Ditto.

linetools/spectra/xspectrum1d.py
```diff
@@ -864,39 +876,42 @@ def write_to_fits(self, outfil, select=False, clobber=True):
         add_wave : bool (False)
           Force writing of wavelengths as array, instead of using FITS
           header keywords to specify a wcs.
+        fill_val : float, optional
+          Fill value
```
ntejos: Fill value for masked pixels.

linetools/spectra/xspectrum1d.py
ntejos: I mean, this may have implications for plotting...

linetools/spectra/xspectrum1d.py
```diff
@@ -182,6 +184,12 @@ def __init__(self, wave, flux, sig=None, co=None, units=None, select=0,
             if wave.shape[0] != sig.shape[0]:
                 raise IOError("Shape of wave and sig vectors must be identical")
             self.data['sig'] = np.reshape(sig, (self.nspec, self.npix))
+            if mask_edges:
+                for kk in range(self.nspec):
+                    sigval = np.where(self.data['sig'][kk] > 0.)[0]
+                    for key in self.data.dtype.names:
+                        self.data[key][kk][0:sigval[0]].mask = True
+                        self.data[key][kk][sigval[-1]+1:].mask = True
         else:
```
ntejos: If sigma is None, then how do we mask?

ntejos:
Looks good to me; please consider the comments left before merging.

profxj:
These are fine suggestions, but this 'mask_edges' is definitely the way I would
prefer to mask.  For plotting and other reasons.  Am happy to add in other methods
as well including a fill_value schema.

But the starting point, I think is a masked array.  So that is the main issue here.
Will wait for @nhmc to return before I proceed.

nhmc:
I've had bad experiences with masked arrays in the past. Doing operations on them can be slow and they often make things unnecessarily complicated (although things may have improved since I last used them).

For cases like this I set the flux and error for bad pixels to np.nan and then make sure the rest of the code can handle that gracefully.  Would doing that work in this case?

profxj:
Good points.  I do prefer that we really do want a mask here.  But let
me fiddle about with NANs and see what I see.

profxj:
Maybe I'm not trying these right, but changing them to nan's
is not enough.  I presume I need to then do filtering.  Trying..

linetools/spectra/xspectrum1d.py
profxj: It does not.  Hence the named 'edges'.  
I'm adding in a mask_all which will do all sig<=0

linetools/spectra/xspectrum1d.py
profxj: fixed

linetools/spectra/xspectrum1d.py
profxj: Now is the option  masking='all'

linetools/spectra/xspectrum1d.py
profxj: definitnely not a good approach.  0. is a value

linetools/spectra/xspectrum1d.py
profxj: added

linetools/spectra/xspectrum1d.py
profxj: We do not mask

linetools/spectra/xspectrum1d.py
```diff
@@ -182,6 +192,15 @@ def __init__(self, wave, flux, sig=None, co=None, units=None, select=0,
...
             self.data['sig'] = np.reshape(sig, (self.nspec, self.npix))
+            if masking != 'None':
+                for kk in range(self.nspec):
+                    sigval = np.where(self.data['sig'][kk] > 0.)[0]
+                    for key in self.data.dtype.names:
+                        if masking == 'edges':
+                            self.data[key][kk][0:sigval[0]].mask = True
+                            self.data[key][kk][sigval[-1]+1:].mask = True
+                        elif masking == 'all':
+                            self.data[key][kk][sigval].mask = True
```
ntejos: This approach will mask out the good pixels and leave in the bad ones. 'True' -> 'False'

profxj: Yikes!  Good catch. 

On June 20, 2016 8:26:50 AM PDT, Nicolas Tejos <notifications@github.com> wrote:
>> @@ -182,6 +192,15 @@ def __init__(self, wave, flux, sig=None,
>co=None, units=None, select=0,
>>              if wave.shape[0] != sig.shape[0]:
>>                  raise IOError("Shape of wave and sig vectors must be
>identical")
>>              self.data['sig'] = np.reshape(sig, (self.nspec,
>self.npix))
>> +            if masking != 'None':
>> +                for kk in range(self.nspec):
>> +                    sigval = np.where(self.data['sig'][kk] > 0.)[0]
>> +                    for key in self.data.dtype.names:
>> +                        if masking == 'edges':
>> +                            self.data[key][kk][0:sigval[0]].mask =
>True
>> +                            self.data[key][kk][sigval[-1]+1:].mask =
>True
>> +                        elif masking == 'all':
>> +                            self.data[key][kk][sigval].mask = True
>
>This approach will mask out the good pixels and leave in the bad ones.
>'True' -> 'False'
>
>---
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/linetools/linetools/pull/226/files/cd9c6988891be9cf2fd10d75db2642452c237954#r67709919
>
>!DSPAM:826,57680b3c29679473325066!

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #227, opened 2016-06-18
Add z to lt_line

profxj:
Adds an option to show redshift lines in lt_line

As per @astronomeara request
@astronomeaera is pleased


astronomeara:
Thanks for the change, merging

PR #231, opened 2016-06-20
Xspec nans

profxj:
Here is another approach using nans.  I still find myself forced
to use masks.  And I feel safer using numpy masked arrays then
my own.  But maybe there is another option?

Please advise


nhmc:
Ok, thanks for doing this. 

If the masking is easier (which it looks like it is), then go with your original PR. Please merge when you're ready.

profxj:
I suppose it is about even as far as 'easy'.  Maybe easier with the 2nd approach.
It's just that my gut tells me to go with numpy then my own approach.

If it burns me, I know how to run..
Am going to sleep on it and choose one.

PR #234, opened 2016-06-22
Stitch and much more

profxj:
This started as a new method to stich spectra together.

It grew into a significant refactoring due to the masked arrays.
Not sure I am regretting the change, but this was a bit painful..

Docs for XSpectrum1D need updating.  Didn't have the endurance
to do that.

New and many revised tests

### DIFF
--- a/linetools/analysis/tests/test_continuum.py
+++ b/linetools/analysis/tests/test_continuum.py
@@ -5,10 +5,11 @@ from ...spectra.io import readspec
 from ..continuum import find_continuum
 import imp
 import numpy as np
+import pytest
 
 def test_find_continuum():
     d = imp.find_module('linetools')[1]
-    spec = readspec(d + '/spectra/tests/files/q0002m422.txt.gz')
+    spec = readspec(d + '/spectra/tests/files/q0002m422.txt.gz', masking='none')
     co, pts = find_continuum(spec, redshift=2.76, divmult=3.5,
                        forest_divmult=3, kind='QSO')
     assert np.allclose(co[:3],
--- a/linetools/spectra/io.py
+++ b/linetools/spectra/io.py
@@ -23,7 +23,6 @@ from astropy.io.fits.hdu.table import BinTableHDU
 
 from .xspectrum1d import XSpectrum1D
 
-#from xastropy.xutils import xdebug as xdb
 
 def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
              format='ascii', exten=None, debug=False, select=0, **kwargs):
@@ -77,6 +76,8 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
             if chk == 0:
                 raise IOError('File does not exist {}'.format(specfil))
             hdulist = fits.open(os.path.expanduser(datfil))
+        elif '.hdf5' in specfil:  # HDF5
+            return parse_hdf5(specfil)
         else: #ASCII
             tbl = Table.read(specfil,format=format)
             # No header?
@@ -191,7 +192,10 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
                 xspec1d.data['co'] = fits.getdata(co_filename)
 
     # Add in the header
-    xspec1d.head = head0
+    #xspec1d.head = head0
+    xspec1d.meta['headers'][0] = head0
+    if xspec1d.nspec > 1:
+        warnings.warn("Read in only 1 header (into meta['headers'][0]")
 
     # Return
     return xspec1d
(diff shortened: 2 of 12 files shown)


linetools/analysis/tests/test_continuum.py
```diff
@@ -5,10 +5,11 @@
 from ..continuum import find_continuum
 import imp
 import numpy as np
+import pytest
```
nhmc: Is this import needed?

linetools/spectra/utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+def splice_two(spec1, spec2, wvmx=None, scale=1., chk_units=True):
+    """ Combine two overlapping spectra.
+
+    Parameters
+    ----------
+    spec1 : XSpectrum1D
+    spec2 : XSpectrum1D
+      The overlapping spectrum. It should cover wavelengths
+      *longer* than the original spectrum.
```
nhmc: Can you add a check for this in the code below? (i.e. spec1.wv[-1] > spec2.wv[0] and spec1.wv[0] < spec2.wv[-1] )

linetools/spectra/xspectrum1d.py
```diff
@@ -71,6 +80,51 @@ def from_spec1d(cls, spec1d):
         return slf
 
     @classmethod
+    def from_list(cls, xspecs, **kwargs):
+        """ Generate from a list of XSpectrum1D
+        By default, the excess pixels are masked
+        If sig or co is set in one, it should be set in all...
+        Units must be identical throughout too
+
+        Also note that masked pixels will be lost!
```
nhmc: Can you make this explanation more clear? 

Something like "Generate a single XSpectrum1D instance containing an array of spectra from a list of individual XSpectrum1D spectra. Each spectrum is padded with extra pixels so that the wavelength ranges of all spectra are covered. Padded pixels are masked." (?)

nhmc:
Apart from the inline comments, this looks good to me. Note that we should really start adding a line to  the CHANGES.rst file for big changes like this. (And for every PR, ideally!)

linetools/analysis/tests/test_continuum.py
profxj: Not really, but it doesn't hurt and 
pytest.set_trace() is the same as pdb.set_trace() 
so it replaces an (unnecessary) import pdb.

linetools/spectra/utils.py
profxj: I've added a simpler check:

    '''if spec2.wvmax < spec1.wvmax:
        raise IOError("The second input spectrum does not cover longer wavelengths.")'''

linetools/spectra/xspectrum1d.py
profxj: Updated.

profxj:
Edited CHANGES.rst (a bit) and will continue to do so from now on.

Got carried away and added hdf5 I/O.  Please review.
Unfortunately, this adds a new dependency (h5py).

Please review these new changes before merging..

linetools/spectra/io.py
```diff
@@ -538,6 +542,57 @@ def parse_linetools_spectrum_format(hdulist):
...
+    """
+    import json
+    import h5py
+    # Open
+    if isinstance(inp, basestring):
+        hdf5 = h5py.File(inp, 'r')
+    else:
+        hdf5 = inp
+    # Data
+    data = hdf5['data'].value
```
nhmc: It seems more natural to me to have `hdf['flux']` and `hdf['wavelength']`, instead of `hdf['data']` containing both.

linetools/spectra/io.py
```diff
@@ -538,6 +542,57 @@ def parse_linetools_spectrum_format(hdulist):
...
+    # Meta
+    if 'meta' in hdf5.keys():
```
nhmc: Attributes are the official way to store metadata in hdf5 (http://docs.h5py.org/en/latest/quick.html#attributes) 

It would be better to use these instead of dumping the json string.

nhmc:
I left some comments about the new hdf5 format. Can you use attributes to store the meta data instead of a dedicated data field? (Maybe not for some reason...)

linetools/spectra/io.py
profxj: data is a straightforward writing of the masked numpy array
(see XSpectrum1D).  As such, this is the way to read it.

linetools/spectra/io.py
profxj: Check out the write_to_hdf5 method in XSpectrum1D
and the meta_to_disk mtehod in spectra.utils.

This has gotten a bit complex to handle multiple header
files for multiple spectra.

nhmc: Ok, I see now. This seems fine then.

linetools/spectra/io.py
nhmc: Oh right, I actually mis-read the code here.

nhmc:
Looks good, please merge when ready.

profxj:
merging

PR #235, opened 2016-06-23
xspec: plot filename of spectrum (if any).

ntejos:
As stated...


profxj:
I feel like this is done in my pending PR, but I merge anyhow.

PR #236, opened 2016-06-23
minor tweaks for speed up in abssystem

ntejos:
As stated...

### DIFF
--- a/linetools/isgm/abssystem.py
+++ b/linetools/isgm/abssystem.py
@@ -27,7 +27,7 @@ from linetools.spectralline import AbsLine
 from linetools.abund import ions
 
 # Globals to speed things up
-c_mks = const.c.to('km/s')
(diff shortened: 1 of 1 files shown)


profxj:
Can you keep c_mks as originally defined (I believe we use
it this way throughout linetools) and just add a new global, 
e.g.  c_mks_value ?

ntejos:
Let me check what was the previous convention. I thought that it was `c_mks` as a float number in MKS units...

ntejos:
It seems that this is the only file within linetools where we have used this variable. The only other time that we have used it (that I recall) was in IGMGuesses, and the convention was the one I just adopted in this PR. Given that the name already include the unit system (i.e. MKS), it does not make much sense to have it as a `Quantity`...

profxj:
ok, check and merge.

PR #237, opened 2016-06-24
Info in the IO

ntejos:
More tags for IO inputs, and some extra info and error regarding reading ASCII files.


profxj:
1. Add tests to keep the coverage as high
2. Update CHANGES.rst
3. Examine PR #234 which has new I/O and see if there is any conflict

ntejos:
Will wait until PR #234 is merged then ;)

profxj:
Ok, merge with master and see what you've got.  Thanks for your patience.

PR #239, opened 2016-06-30
fit continuum: fixed to handle the masked arrays.

ntejos:
As stated. I still do not like much representing the flux, sigma attributed using MaskedArray.compressed()...

### DIFF
--- a/linetools/spectra/xspectrum1d.py
+++ b/linetools/spectra/xspectrum1d.py
@@ -219,13 +219,13 @@ class XSpectrum1D(object):
         """
         # Error checking
         if not isinstance(wave, np.ndarray):
-            raise IOError("Input x-vector must be an ndarray")
+            raise IOError("Input `wave` vector must be an numpy.ndarray.")
         if not isinstance(flux, np.ndarray):
-            raise IOError("Input y-vector must be an ndarray")
+            raise IOError("Input `flux` vector must be an numpy.ndarray.")
         if wave.shape[0] != flux.shape[0]:
-            raise IOError("Shape of x and y vectors must be identical")
+            raise IOError("Shape of `flux` and `wave` vectors must be identical.")
         if masking not in ['none', 'edges', 'all']:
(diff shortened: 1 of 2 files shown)


profxj:
you can set masking='none'  :)

Merge when this passes the test.

profxj:
I should add that it is better for one to *not* reassign the
internal data of an XSpectrum1D.  Best to create a new object.

linetools/spectra/xspectrum1d.py
```diff
@@ -1301,15 +1304,17 @@ def fit_continuum(self, knots=None, edges=None, wlim=None, dw=10.,
...
                                    contpoints, co=co_init, fig=fig, anchor=anchor)
 
         # wait until the interactive fitting has finished
         while not wrapper.finished:
             plt.waitforbuttonpress()
 
         print('Updating continuum')
-        self.data['co'] = wrapper.continuum
+        # import pdb; pdb.set_trace()
+        gdp = ~self.data['wave'][self.select].mask
```
ntejos: I see, so this will crash when `masking='none'`...

profxj: I don't think it should.  Will it?

ntejos: Not so sure anymore. Will wait until tomorrow morning to give it a second pass before merging then (and to solve the Python 3 error).

ntejos:
Regarding reassignation of internal self.data, how would it work when using the internal method fit_continuum? Would you prefer self.fit_continuum() to return a copy of itself with the continuum?

ntejos:
@profxj Do you mind if I merge?

profxj:
Internal assignment can be done.  I see now that it was simply
being done wrong in that code.  

For continua, you can do:

self.co = xxxx

but if it is the first time you are setting it (as I presume for fit_continuum)
then you should use the normalize() method.

Consider htis then merge.

PR #245, opened 2016-07-09
HDF5 script

profxj:
New script for viewing HDF5 files.

Not really a linetools necessity, but 
quite useful.

Still failing Python 3.5 issues


profxj:
This is more or less handled by h5ls and h5dump.

Closing.

PR #246, opened 2016-07-10
Absline Kinematics

profxj:
New class for kinematic analysis
Tests but no new docs

Also fixes Python 3.5 test failures


ntejos:
Why the need of a new Class instead of being incorporated as methods of `AbsLine`?

profxj:
A fair question.  I found it easier to package everything
this way but I have actually then simply grabbed the data dict and
ignored the object (in COS-Halos/COS-Dwarfs).

profxj:
Shall I reconsider (again)?

mneeleman:
I actually don't have a strong opinion either way. I might lean a bit towards Nicolas's point of view, as it seems to make sense to treat the kinematic analysis in the same way as the abundance measurements (e.g., via the AODM method).

That said, I looked at the code and all seems to work for me, except for the u'v_peak' measurement, which seem to give wacky values. This part of the code (line 177-178) 
doesn't make sense to me:

        imx = np.argmax(self.stau)
        self.data['v_peak'] = self.velo[self.pix[imx]]

profxj:
Ok, to methods we go.
I've left the class in, but can remove it at a later date (or now if you prefer).
@ntejos please merge if you like it.

ntejos:
I would prefer to remove the Class before merging, so the code remains cleaner.

profxj:
Please do so

On July 16, 2016 8:28:04 AM PDT, Nicolas Tejos <notifications@github.com> wrote:
>I would prefer to remove the Class before merging, so the code remains
>cleaner.
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/linetools/linetools/pull/246#issuecomment-233135590
>
>!DSPAM:826,578a529133702987917289!

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

profxj:
Merging.

PR #248, opened 2016-07-14
Air to vac

profxj:
air to vac method in xspectrum1D
test
added option to lt_xspec

### DIFF
--- a/linetools/spectra/xspectrum1d.py
+++ b/linetools/spectra/xspectrum1d.py
@@ -517,6 +517,29 @@ class XSpectrum1D(object):
         #
         return newspec
 
+    def airtovac(self):
+        """ Converts current wavelength array from an assumed air to vacuum wavelength scale
+
+        Returns
+        -------
+        Resultant wavelength array is in AA no matter the input units
+        """
+        # Convert to AA
+        wavelength = self.wavelength.to(u.AA).value
+
+        # Standard conversion format
+        sigma_sq = (1.e4/wavelength)**2. #wavenumber squared
+        factor = 1 + (5.792105e-2/(238.0185-sigma_sq)) + (1.67918e-3/(57.362-sigma_sq))
+        factor = factor*(wavelength>=2000.) + 1.*(wavelength<2000.) #only modify above 2000A
+
+        # Convert
+        wavelength = wavelength*factor
+        # Units
+        new_wave = wavelength*u.AA
+        # Finish
+        self.wavelength = new_wave
+
+
     def constant_sig(self, sigv=0.):
         """Set the uncertainty array to a constant value.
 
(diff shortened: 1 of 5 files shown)


linetools/spectra/xspectrum1d.py
```diff
@@ -517,6 +517,29 @@ def add_noise(self, seed=None, s2n=None, rstate=None):
         #
         return newspec
 
+    def airtovac(self):
+        """ Converts current wavelength array from an assumed air to vacuum wavelength scale
+
+        Returns
+        -------
+        Resultant wavelength array is in AA no matter the input units
```
ntejos: typo

linetools/spectra/xspectrum1d.py
```diff
@@ -517,6 +517,29 @@ def add_noise(self, seed=None, s2n=None, rstate=None):
...
+        factor = factor*(wavelength>=2000.) + 1.*(wavelength<2000.) #only modify above 2000A
+
+        # Convert
+        wavelength = wavelength*factor
+        # Units
+        new_wave = wavelength*u.AA
+        # Finish
+        self.wavelength = new_wave
+
+
```
ntejos: I think it would be nice to have the complementary function vactoair().

linetools/spectra/xspectrum1d.py
profxj: I don't see it.

linetools/spectra/xspectrum1d.py
profxj: Added, although not to lt_xspec (vac is default)
Test too

linetools/spectra/xspectrum1d.py
ntejos: Resultant

linetools/spectra/xspectrum1d.py
```diff
@@ -517,6 +517,50 @@ def add_noise(self, seed=None, s2n=None, rstate=None):
...
+        factor = factor*(wavelength>=2000.) + 1.*(wavelength<2000.) #only modify above 2000A
+
+        # Convert
+        wavelength = wavelength/factor
+        # Units
+        new_wave = wavelength*u.AA
+        # Finish
+        self.wavelength = new_wave
+
     def constant_sig(self, sigv=0.):
```
ntejos: Finally, it would be nice to have a way to prevent the user from applying these convertions more than once. i.e. imagine that the spectrum is already calibrated at vacuum and airtovac() is applied again that would be problematic. Maybe setting an attribute like self.wvcalib = 'air'/'vacuum' ?

profxj: Good suggestion.
Now in meta with 'vac' as default

profxj:
merging

PR #250, opened 2016-07-14
Dot lbl

profxj:
Prints out x,y position for dot in spec_widget.xspec


profxj:
Tested and approved.

Merging

PR #252, opened 2016-07-17
Two small fixes

profxj:
One to avoid redoing the extra_columns
One related to kinematics

### DIFF
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -61,9 +61,7 @@ class LineList(object):
         transitions, rather than from the original source files.
 
     sort_by : str or list of str, optional
-        Keys to sort the underlying data table by.
-        Turning this on all adds the extra columns to the datatable
-
+        Keys to sort the underlying data table by. Default is 'wrest'
     """
 
     # Init
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -61,7 +61,8 @@ class LineList(object):
         transitions, rather than from the original source files.
 
     sort_by : str or list of str, optional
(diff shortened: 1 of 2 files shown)


linetools/lists/linelist.py
```diff
@@ -61,7 +61,8 @@ class LineList(object):
         transitions, rather than from the original source files.
 
     sort_by : str or list of str, optional
-        Keys to sort the underlying data table by. Defauls is 'wrest'.
+        Keys to sort the underlying data table by.
+        Turning this on all adds the extra columns to the datatable
```
ntejos: Keep the default value explicit.

profxj: agreed.  Back in

linetools/lists/linelist.py
```diff
@@ -341,11 +345,16 @@ def set_extra_columns_to_datatable(self, abundance_type='solar', ion_correction=
...
+         redo : bool, optional
+            Remake the extra columns
 
         Note: This function is only implemented for the following
         lists: HI, ISM, EUV, Strong.
 
         """
+        # Avoid redo (especially for caching)
+        if ('ion_name' in self._data.keys()) and (not redo):
+            return
```
ntejos: Looks like a good solution for now, but I think this should be controled at the LineList.__init__() level. I am envisioning this method to set different kinds of abundances, and ionization correction options as given in the initialization of LineList.

An alternative then would be to add a keyword attribute in .__init__() like `set_extra_columns=False` or something like that...

profxj: Yeah, I almost went that route...
I'll leave that refactor to you.  :)

Please merge if you are ok with it.

PR #254, opened 2016-07-18
A few changes for X-Shooter

profxj:
a few changes for X-Shooter data (as provided by ESO)

Also now insists on using Ang in GUIs


ntejos:
Looks good to me. Can you add a test to  `get_wave_unit()`?

ntejos:
Looks good, please merge when ready...

profxj:
Merging

Issue #255, opened 2016-07-18
splice with lt_xspec


ktirimba:
please add the ability to splice two or more spectra in a fits file

profxj:
Done.

PR #256, opened 2016-07-19
Splicing and extensions

profxj:
New code for splicing with lt_xspec

Also enables extension specification in filename suffix

with revised docs.


profxj:
Addresses Issue #255

PR #257, opened 2016-07-20
Abscomponent: overlapping components

ntejos:
Utils for checking whether two components overlap in wobs space.
Tests added accordingly.


linetools/isgm/abscomponent.py
```diff
@@ -238,11 +241,11 @@ def add_absline(self, absline, tol=0.1*u.arcsec, chk_vel=True,
         testE = bool(self.Ej == absline.data['Ej'])
         # Now redshift/velocity
         if chk_vel:
-            dz_toler = (1+self.zcomp)*vtoler/3e5  # Avoid Quantity for speed
-            zlim_line = (1+absline.attrib['z'])*absline.analy['vlim']/const.c.to('km/s')
-            zlim_comp = (1+self.zcomp)*self.vlim/const.c.to('km/s')
-            testv = (zlim_line[0] >= (zlim_comp[0]-dz_toler)) & (
-                zlim_line[1] <= (zlim_comp[1]+dz_toler))
+            dz_toler = (1 + self.zcomp) * np.fabs(vtoler) / c_kms  # Avoid Quantity for speed
```
profxj: Why the fabs on vtoler?  Should it ever be negative?

linetools/isgm/tests/test_use_abscomp.py
```diff
@@ -207,5 +207,16 @@ def test_repr_alis():
...
-    s = abscomp.repr_alis(tie_strs=('a', 'b', 'CD',''),fix_strs=('', 'f', '', ''))
+    s = abscomp.repr_alis(tie_strs=('a', 'b', 'CD',''), fix_strs=('', 'f', '', ''))
     assert s == 'voigt   ion=28Si_II 0.00a redshift=2.92939F 0.0cd 1.0E+04# Something\n'
+
+
+def test_overlapping_components():
+    abscomp, HIlines = mk_comp('HI')
+    SiIIcomp1,_ = mk_comp('SiII',vlim=[50.,300.]*u.km/u.s)
+    SiIIcomp2,_ = mk_comp('SiII',vlim=[-300.,0.]*u.km/u.s)
+    assert ltiu.overlapping_components(abscomp, abscomp)  # should overlap
```
profxj: You might specify z here as well just to be crystal clear.

linetools/isgm/tests/test_use_abscomp.py
```diff
@@ -207,5 +207,16 @@ def test_repr_alis():
...
     s = abscomp.repr_alis()
     assert s == 'voigt   ion=28Si_II 0.00 redshift=2.92939 0.0 1.0E+04# Something\n'
-    s = abscomp.repr_alis(tie_strs=('a', 'b', 'CD',''),fix_strs=('', 'f', '', ''))
+    s = abscomp.repr_alis(tie_strs=('a', 'b', 'CD',''), fix_strs=('', 'f', '', ''))
     assert s == 'voigt   ion=28Si_II 0.00a redshift=2.92939F 0.0cd 1.0E+04# Something\n'
+
+
+def test_overlapping_components():
+    abscomp, HIlines = mk_comp('HI')
+    SiIIcomp1,_ = mk_comp('SiII',vlim=[50.,300.]*u.km/u.s)
```
profxj: You should specify vlim for HI just so it is obvious what the test is doing (i.e. I can't figure it out)

linetools/isgm/tests/test_use_abscomp.py
```diff
@@ -207,5 +207,16 @@ def test_repr_alis():
...
+    SiIIcomp2,_ = mk_comp('SiII',vlim=[-300.,0.]*u.km/u.s)
+    assert ltiu.overlapping_components(abscomp, abscomp)  # should overlap
+    f = ltiu.overlapping_components(abscomp, SiIIcomp1)  # should not overlap
+    assert ~f
+    f = ltiu.overlapping_components(SiIIcomp2, SiIIcomp1) # should not overlap
+    assert ~f
```
profxj: can you add a test which *does* overlap?

linetools/isgm/tests/test_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+# TEST_UNICODE_LITERALS
+
+from linetools.isgm import utils as ltiu
+import astropy.units as u
+import numpy as np
+
+import imp
+lt_path = imp.find_module('linetools')[1]
+
```
profxj: Is this needed?

linetools/isgm/utils.py
```diff
@@ -32,16 +34,16 @@ def chk_components(components, chk_match=False, chk_A_none=False, toler=0.2*u.ar
       if True, require that the components match in RA/DEC, Zion, Ej, A but not velocity
     chk_A_none : bool, optional
       if True, require that A *not* be set
-    toler : Angle, optional
-      Tolerance on matching coordinates
+    tol : Quantity, optional
+      Tolerance on matching SkyCoordinates. Default is 0.2*u.arcsec
```
profxj: SkyCoord

linetools/isgm/utils.py
```diff
@@ -362,3 +366,138 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+    """True if there is overlap between chunks
+    `chunk1` and `chunk2`. Otherwise False. Chunks are
+    assumed to represent continuous coverage, so the only
+    information that matters are the minimum and maximum
+    values of a given chunk.
+
+    Parameters
+    ----------
+    chunk1 : tuple, list, 1-d np.array, Quantity
+        A given chunk, assumed to represent a contiguous region
```
profxj: Does it need to be iterable or can it be a single number (as you suggest with Quantity)?

If iterable, I'd check it with astropy.utils.misc.isiterable

linetools/isgm/utils.py
```diff
@@ -362,3 +366,138 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+        So only its minimum and maximum values matter
+        Still, chunk must be sorted.
+    chunk1 : tuple, list, 1-d np.array
```
profxj: chunk2

linetools/isgm/utils.py
```diff
@@ -362,3 +366,138 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+    assumed to represent continuous coverage, so the only
+    information that matters are the minimum and maximum
+    values of a given chunk.
+
+    Parameters
+    ----------
+    chunk1 : tuple, list, 1-d np.array, Quantity
+        A given chunk, assumed to represent a contiguous region
+        So only its minimum and maximum values matter
+        Still, chunk must be sorted.
```
profxj: I didn't catch the 'must be sorted' on my first read.  Might see if you can make that more prominent.

linetools/isgm/utils.py
```diff
@@ -362,3 +366,138 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+
+    # now the chunk1 is the one with the lowest value
+    # then, the only way they do not overlap is:
+    if np.min(chunk2) > np.max(chunk1):
+        return False
+    else:  # overlap exists
+        return True
+
+def overlapping_components(comp1, comp2, tol=0.2*u.arcsec):
+    """Whether two components overlap in wavelength
```
profxj: 0.2 is pretty tight.  But I suspect I chose that myself.

profxj: Maybe call this 'coincident_components' ?

When I read overlapping, I think there is a physical association, i.e. the gas is around the same galaxy.  But I think you mainly have two components at completely different redshifts in mind.

linetools/isgm/utils.py
```diff
@@ -362,3 +366,138 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+    wobs_chunks_2 = []
+    for absline in comp1._abslines:
+        cond = absline.analy['vlim'] != init_analy['vlim']  # default value?
+        if np.sum(cond) > 0:  # i.e. absline vlim not equal than the default
+            dzlim = give_dz(absline.analy['vlim'], comp1.zcomp)
+        else:  # use the vlim from component otherwise
+            dzlim = give_dz(comp1.vlim, comp1.zcomp)
+        wrest_aux = absline.wrest.to('AA').value  # in AA
+        wobs_aux = wrest_aux * (1 + comp1.zcomp + dzlim)
+        wobs_chunks_1 += [tuple(wobs_aux)]
```
profxj: Ah, I see this is even more complex, i.e. you are considering all the transitions in a given component.  I had though it would be only 2 lines (but of course, I was wrong).

So, I wonder whether there should be a simpler 'coincident_absline' method.  I suppose not.

But I do then wonder why you would care whether 2 components overlap at some 'random' pair of transitions.  That is, I can't see why I'd ever call this method..  Do tell.

linetools/isgm/utils.py
```diff
@@ -362,3 +366,138 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+        else:  # use the vlim from component otherwise
+            dzlim = give_dz(comp2.vlim, comp2.zcomp)
+        wrest_aux = absline.wrest.to('AA').value  # in AA
+        wobs_aux = wrest_aux * (1 + comp2.zcomp + dzlim)
+        wobs_chunks_2 += [tuple(wobs_aux)]
+
+    # now we have two lists of wobs chunks:
+    # so lets do the checking until at least 1
+    # overlap is found
+    for ii in range(len(wobs_chunks_1)):
```
profxj: Ugly for loop.

If you are just going to do this then I do think you should reduce it to a
coincident_absline method and simply loop over the abs_lines as you are doing here instead of chunks.

Yes, I am a pain.

linetools/utils.py
```diff
@@ -8,7 +8,7 @@
 import pdb
 
 import numpy as np
-from astropy import constants as const
+from astropy.constants import c
```
profxj: why this change?

linetools/utils.py
```diff
@@ -272,11 +271,86 @@ def z_from_v(z, v):
...
+    """Gives redshift difference for a given
+    velocity difference(s) at zmean.
+
+    Parameters
+    ---------
+    dv : Quantity or Quantity array
+        Rest-frame velocity at zmean to calculate
+        the corresponding redshift difference, dz
+    zmean : float or np.array
+        Rest-frame redshift to perform the calculation.
```
profxj: Probably is not right to call it 'rest-frame redshift'

profxj:
Hit you with a bunch of inline comments to stew on.

linetools/isgm/abscomponent.py
ntejos: It shouldn't be negative, but the user may enter a negative value... maybe safer to generate an error if the vtoler is negative?

linetools/isgm/tests/test_use_abscomp.py
ntejos: Do you mean in `overlapping_components()`? It should always take the redshift of each component...

linetools/isgm/tests/test_use_abscomp.py
ntejos: Sure.

linetools/isgm/tests/test_use_abscomp.py
ntejos: Line 218 should give you overlap as we are comparing an abscomponent with itself.

linetools/isgm/utils.py
ntejos: Will correct.

linetools/isgm/utils.py
ntejos: The iterability is not the important aspect, but the minimum and maximum value of each chunk.

linetools/isgm/utils.py
ntejos: Thanks.

linetools/isgm/utils.py
ntejos: I added the warning to the main description too.

linetools/isgm/utils.py
ntejos: Indeed, you chose 0.2 arcsec... if we change it here, we should change it all across linetools for consistency...

Will think about a better name... not sure about coincident_components either. How about superimposed_components? (I still prefer overlapping but if in english it may be understood as associated I agree we should find a different name for this).

linetools/isgm/utils.py
ntejos: I kind of like the idea of having the equivalent method for individual abslines and will probably do it that way.

The idea is to identify components whose abslines share the same part of a given spectra, so the fit must be done simultaneously. It is written in the main docstring of the method, but may need extra explanation.

linetools/isgm/utils.py
ntejos: I agree with this suggestion, and will do so. Thanks!

linetools/utils.py
ntejos: Just my preference of reducing the imports...

linetools/utils.py
ntejos: Agreed. Will change it.

ntejos:
Thanks X for your comments. Will work on creating the equivalent method for abslines and build the component one from there. Please do not merge for now.

linetools/isgm/utils.py
profxj: Overlapping doesn't demand physical assocation but it is what I think of.
Why not coincident?

linetools/isgm/utils.py
profxj: Ok, got it.

profxj:
Roger that.  Standing by

ntejos:
@profxj this PR is ready for review again. Thanks!

linetools/isgm/tests/test_use_abscomp.py
```diff
@@ -207,5 +243,48 @@ def test_repr_alis():
...
+    abscomp, HIlines = mk_comp('HI', zcomp=2.92939)
+    SiIIcomp1,_ = mk_comp('SiII',vlim=[50.,300.]*u.km/u.s, zcomp=2.92939)
+    SiIIcomp2,_ = mk_comp('SiII',vlim=[-300.,0.]*u.km/u.s, zcomp=2.92939)
+    assert ltiu.coincident_components(abscomp, abscomp)  # should overlap
+    assert not ltiu.coincident_components(abscomp, SiIIcomp1)  # should not overlap
+    assert not ltiu.coincident_components(SiIIcomp2, SiIIcomp1) # should not overlap
+    with pytest.raises(ValueError):
+        a = ltiu.coincident_components('not_a_component', SiIIcomp1)
+    with pytest.raises(ValueError):
+        a = ltiu.coincident_components(abscomp, 'not_a_component')
```
profxj: That is one hell of a test suite!

linetools/isgm/utils.py
```diff
@@ -362,3 +366,89 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+        raise ValueError('comp1 must be AbsComponent object.')
+
+    # Check whether they are in the same sky region
+    if comp1.coord.separation(comp2.coord) > tol:
+        return False
+
+    # loop over abslines
+    for line1 in comp1._abslines:
+        for line2 in comp2._abslines:
+            overlap = line1.coincident_lines(line2)
```
profxj: coincident_line instead of coincident_lines ?

linetools/spectralline.py
```diff
@@ -386,6 +386,29 @@ def to_dict(self):
         # Return
         return adict
 
+    def coincident_lines(self, specline):
```
profxj: Rename to coincident_line

linetools/utils.py
profxj: I think mine is the standard use in astropy

profxj:
Added a few more nits.  Consider/address those
and merge.

linetools/spectralline.py
ntejos: Sure.

linetools/utils.py
ntejos: Ok, will come back to original...

PR #258, opened 2016-07-20
Add flux error to lt_xspec

ktirimba:
Modifies spec_widget to propoagate and print flux error

Tested by Tirimba on DEIMOS data


profxj:
Merging

PR #261, opened 2016-07-24
q is for Quit

profxj:
Add 'q' as the key to quit lt_xspec
Addresses Issue#243

Also fixes a dumb bug

Tested by JXP


astronomeara:
verified that 'q' works as advertised!

profxj:
merging

PR #262, opened 2016-07-25
Abscomp utils: group_coincident_components()

ntejos:
- group_coincident_components() implemented
- with test

This function will be useful for VP fitting a full spectrum, where components that are isolated are easy to identify, and so those that are coincident in wobs with others. It may be a bit of a pain to review it...

### DIFF
--- a/linetools/isgm/utils.py
+++ b/linetools/isgm/utils.py
@@ -516,15 +516,21 @@ def group_coincident_compoments(comp_list):
                 pass
 
     # Now we have out as a list of lists with indices, or empty lists
-    # So lets produce the final output from it
+    # let's get rid of the empty lists
+    out = [x for x in out if x != []]
+
+    # Now lets produce the final output from it
     output_list = []
+    output_dict = {}
     for ii in range(len(out)):
-        if len(out[ii]) == 0:
-            continue
         aux_list = []
         for jj in out[ii]:
             aux_list += [comp_list[jj]]
+        output_dict['{}'.format(ii)] = aux_list
         output_list += [aux_list]
+
+    # choose between dict of list
+    # return output_dict
     return output_list
 
 
--- a/linetools/isgm/utils.py
+++ b/linetools/isgm/utils.py
@@ -475,7 +475,7 @@ def group_coincident_compoments(comp_list, output_type='list'):
         in the output.
     """
     if output_type not in ['list', 'dict', 'dictionary']:
-        raise ValueError("`output_type` must be either 'list' or 'dict'".)
+        raise ValueError("`output_type` must be either 'list' or 'dict'.")
 
     # the first extreme case is that all components are independent
     # of each other, in which case we have the following output shape
--- a/linetools/isgm/utils.py
+++ b/linetools/isgm/utils.py
@@ -455,9 +455,9 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
 
 
 def group_coincident_compoments(comp_list):
-    """For a given input list of component, this function
-    group together components that are coincident to each other
-    (including by transitivity), and return them as a list of
+    """For a given input list of components, this function
+    groups together components that are coincident to each other
+    (including by transitivity), and returns them as a list of
     component lists.
 
     Parameters
@@ -491,8 +491,7 @@ def group_coincident_compoments(comp_list):
         for jj in range(ii+1, len(comp_list)):
             # print(ii,jj)
             comp_jj = comp_list[jj]
-            overlap_ii_jj = coincident_components(comp_ii, comp_jj)
-            if overlap_ii_jj is True:
+            if coincident_components(comp_ii, comp_jj):  # There is overlap between comp_ii and comp_jj
                 # check in the previous ones where does jj belongs to
                 switch = 0
                 for kk in range(len(out[:ii])):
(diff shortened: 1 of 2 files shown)


linetools/isgm/utils.py
```diff
@@ -451,4 +451,81 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
...
             if overlap is True:
                 return True
-    return False
\ No newline at end of file
+    return False
+
+
+def group_coincident_compoments(comp_list):
+    """For a given input list of component, this function
+    group together components that are coincident to each other
```
profxj: groups together

linetools/isgm/utils.py
```diff
@@ -451,4 +451,81 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
...
+    (including by transitivity), and return them as a list of
```
profxj: returns them

linetools/isgm/utils.py
```diff
@@ -451,4 +451,81 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
...
+    component lists.
+
+    Parameters
+    ----------
+    comp_list : list of AbsComponent
+        Input list of components to group
+
+    Returns
+    -------
+    output : list of lists of AbsComponent
```
profxj: Could the output be instead a dict of lists?
Where the keys had some (simple) meaning, e.g. mean redshift?

linetools/isgm/utils.py
```diff
@@ -451,4 +451,81 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
...
+        The grouped components as individual lists
+        in the output list.
+    """
+    # the first extreme case is that all components are independent
+    # of each other, in which case we have the following output shape
+    out = [[] for kk in range(len(comp_list))]
+
+    for ii in range(len(comp_list)):
```
profxj: I prefer to write these as:

for ii,comp_ii in enumerate(comp_list):

Just something to consider going forward.

linetools/isgm/utils.py
```diff
@@ -451,4 +451,81 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
...
+        if switch == 1:
+            pass
+        else:
+            out[ii].append(ii)
+
+        for jj in range(ii+1, len(comp_list)):
+            # print(ii,jj)
+            comp_jj = comp_list[jj]
+            overlap_ii_jj = coincident_components(comp_ii, comp_jj)
+            if overlap_ii_jj is True:
```
profxj: Can combine 494 and 495, i.e.

if coincident_components(comp_ii, comp_jj):

linetools/isgm/utils.py
```diff
@@ -451,4 +451,81 @@ def coincident_components(comp1, comp2, tol=0.2*u.arcsec):
...
+    # So lets produce the final output from it
+    output_list = []
+    for ii in range(len(out)):
+        if len(out[ii]) == 0:
+            continue
+        aux_list = []
+        for jj in out[ii]:
+            aux_list += [comp_list[jj]]
+        output_list += [aux_list]
+    return output_list
```
profxj: I still think a dict of lists would be easiest to read
and maybe to debug.   Consider it unless the
ordering in the final product is important and needs
to be preserved.

linetools/isgm/utils.py
ntejos: Thanks.

linetools/isgm/utils.py
ntejos: Thanks.

linetools/isgm/utils.py
ntejos: I thought about it, but I could not come with a sample naming convention for the keys that make sense. Remember that these are components being coincident in wobs space that may or may not be physically associated; mean redshift will be misleading. Because of the lack of a good naming for the keys, I decided to go for an output in list form rather than dictionary.

linetools/isgm/utils.py
ntejos: Do not how to do the enumerate for the `jj` index in current line 491, as `jj` has to start from `ii+1`. Will leave it as is for clarity.

linetools/isgm/utils.py
ntejos: Sure.

linetools/isgm/utils.py
ntejos: Ordering is not important, but do you have a suggestion for a meaningful key for the dict?

profxj: I'm suggesting something related to redshift

ntejos: Something related to redshift is misleading because the coincident components may have widely different redshifts (it only matters that any of their individual abslines overlap in wvobs space).

profxj: Well, how about

z###_z###_z###

where ### is the redshift to some precision?

ntejos: Such suggestion wont work because you may have two independent grouped components having the same name, if for some reason they hay the same redshift distribution. If you insist on dict() output I think the best option would be to call the keys: `'0'`, `'1'`, `'2'`... and so on.

profxj:
Merging, although I can tell you that working
with lists of lists makes one's head spin..  ;)

PR #266, opened 2016-07-29
Fix Verner96 bug

profxj:
Addes names to Verner transitions

And a bit more

No new docs or tests


jessicawerk:
This seems fine to me. Seems like it fixed both Verner 94 and Verner 96? I'd add a comment in-line, but I have no idea what to say...

profxj:
Merging.

PR #267, opened 2016-07-30
box_smooth failed if there was no error array and preserve was set to True

banados:
box_smooth failed if there was no error array and preserve was set to True

These lines fix this issue: First checks if an error array exists, if so it does the convolution otherwise returns None for the "new" error array


banados:
noted a mistake, i will close it and open it again

PR #268, opened 2016-07-30
Fixing box_smooth when no error array is given

banados:
box_smooth failed when no error array was given and "preserve" was set to True.

These lines fixed that issue by checking first if there is an error array, and setting the new error array (new_sig) to None if self.sig_is_set == False


profxj:
Good fix.  Merging.

PR #269, opened 2016-07-30
Load ZH

profxj:
Enables ZH loading from JSON dicts
in AbsSystem

Plus a little refactoring..

### DIFF
--- a/linetools/isgm/abssystem.py
+++ b/linetools/isgm/abssystem.py
@@ -153,18 +153,12 @@ class AbsSystem(object):
         return slf
 
     @classmethod
-    def from_dict(cls, idict, skip_components=False, use_coord=False, **kwargs):
+    def from_dict(cls, idict, use_coord=False, **kwargs):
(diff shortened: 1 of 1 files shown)


linetools/isgm/abssystem.py
```diff
@@ -226,6 +208,10 @@ def __init__(self, radec, zabs, vlim, zem=0., abs_type=None,
         # Kinematics
         self.kin = {}
 
+        # Metallicity
```
jessicawerk: The attribute in the full code is called MH, and this is ZH. I can't recall if M/H and Z/H are different quantities (this always confuses me), but I think we should be consistent in the naming.

jessicawerk:
Other than the quibble between MH and ZH, This looks good to me.

linetools/isgm/abssystem.py
profxj: Good catch.  

We go with ZH

profxj:
Merging

PR #271, opened 2016-08-01
Test coverage

ntejos:
Increasing coverage...


profxj:
These look great.

Merge when they pass.

PR #276, opened 2016-08-04
Line widget

ntejos:
New option in select_line_widget adding redshift of line if such keyword (i.e. `'redshift'`) is present in the input table. Useful for IGMGuesses and potentially other GUIs too.


profxj:
Looks fine.  Merging

PR #278, opened 2016-08-05
Joevp repr

ntejos:
Adding convenient functions for passing from components to JOEVP input file.

- I also added `AbsComponent.attrib` as dict() that could be potentially filled out. Happy to reconsider other options.


linetools/isgm/abscomponent.py
```diff
@@ -594,6 +598,49 @@ def repr_alis(self, T_kin=1e4*u.K, bturb=0.*u.km/u.s,
         s += '\n'
         return s
 
+    def repr_joevp(self, specfile, flags=(1,1,1), b_default=10*u.km/u.s):
```
jnburchett: We need to avoid using '1' as a flag, as this value is reserved to denote line parameters that are to be fixed.  The other reserved value is '0', which denotes parameters that may freely vary, i.e., not tied to any other parameter.

linetools/isgm/utils.py
```diff
@@ -528,3 +528,35 @@ def group_coincident_compoments(comp_list, output_type='list'):
         return output_dict
 
 
+def complist_to_joevp(comp_list, specfile, outfile):
+    """ From a given component list, it produces an
+    input file for JOEVP (Voigt profile fitter).
+
```
jnburchett: Note that my code is called (for now) 'joebvp'.  Not to be confused with 'joe_vp', which was written by Joe Meiring.  I really don't care about this, but just thought I'd mention it.  Easily refactored, of course, if you so choose.

linetools/isgm/tests/test_use_abscomp.py
```diff
@@ -252,6 +270,37 @@ def test_repr_alis():
...
+    compare_two_files(data_path('test_joevp_repr.joevp'), data_path('test_joevp_repr_reference.joevp'))
+    # now add attribute to comp and compare again
+    abscomp.attrib['b'] = 15*u.km/u.s
+    ltiu.complist_to_joevp(comp_list, 'test.fits', data_path('test_joevp_repr.joevp'))
+    compare_two_files(data_path('test_joevp_repr.joevp'), data_path('test_joevp_repr_reference.joevp'))
+
+
 def test_get_wvobs_chunks():
     abscomp, HIlines = mk_comp('HI', zcomp=0., vlim=[0,10]*u.km/u.s)
     wvobs_chunks = ltiu.get_wvobs_chunks(abscomp)
```
jnburchett: I don't see the 'get_wvobs_chunks' definition within this PR (or my current version of linetools), but I suspect this returns the transitions for each component.  The output files provided for our test sightline (3C273) include many transitions that are outside the wavelength range covered by the COS spectrum, and I think these files would be easier to parse and edit manually (if needed) if they contained the transitions covered by the spectrum being fitted.  Is this an easy switch to add?   

I understand the reasoning behind the bookkeeping of components being general enough to not specify specific transitions internally, but perhaps we could limit the transitions just shown in these input files.  Todd and I are finding that opening up the parameter files and tweaking them is sometimes necessary.

jnburchett: FYI, joebvp only tries to fit the lines covered in the spectrum, so the extra lines in the input files do not cause any problems from the code's perspective.

linetools/isgm/abscomponent.py
ntejos: Got it, `(2,2,2)` set as default.

linetools/isgm/utils.py
ntejos: Sure, no problem.

linetools/isgm/tests/test_use_abscomp.py
ntejos: Thanks, will think about the best way to implement this...

ntejos:
Not ready for merging yet...

ntejos:
@profxj could you please review the last commit about to_dict() and from_dict()? Also, any suggestion to make the test pass python 3+? There is a discrepancy converting float point to string...

profxj:
on it

linetools/isgm/abscomponent.py
```diff
@@ -594,6 +598,49 @@ def repr_alis(self, T_kin=1e4*u.K, bturb=0.*u.km/u.s,
...
+        # specfile|restwave|zsys|col|bval|vel|nflag|bflag|vflag|vlim1|vlim2|wobs1|wobs2|trans
+        s = ''
+        for aline in self._abslines:
+            s += '{:s}|{}|'.format(specfile, aline.wrest.to('AA').value)
+            logN = aline.attrib['logN']
+            b_val = aline.attrib['b'].to('km/s').value
+            if b_val == 0:  # set the default
+                b_val = b_default.to('km/s').value
+            s += '{}|{}|{}|0.|'.format(self.zcomp, logN, b_val)  # `vel` is set to 0. because z is zcomp
+            s += '{}|{}|{}|'.format(flags[0], flags[1], flags[2])
```
profxj: Use explicit formatting here to avoid the floating point issues in Python 3,e.g.

{:0.4f} instead of {}

profxj:
Ok, I think I've done my duty.  Note my inline comment as
regards the Python 3 issue.

linetools/isgm/tests/test_use_abscomp.py
ntejos: I have finally decided that the way to handle this, will be at the AbsComponent level, i.e. it is the user's responsibility to define AbsComponent with AbsLines that a given spectrum covers. In the case of IGMGuesses, a new PR will address this.

linetools/isgm/abscomponent.py
ntejos: Thanks

linetools/isgm/tests/test_use_abscomp.py
jnburchett: Do you mean the user of IGMGuesses or the user of joebvp?

ntejos: None of the above, I meant the user of linetools. In the case of IGMGuesses, it will be handled internally to get you what you want ;) A new PR (in pyigm) will come soon...

jnburchett: Got it!

ntejos:
@jnburchett This PR is ready for review again; thanks!

PR #279, opened 2016-08-12
LSF: changing normalization of LSF kernels; imposing interpolated neg…

ntejos:
…ative values to be == 0.


ntejos:
Will close this PR for now...

PR #281, opened 2016-08-12
More LSF

ntejos:
- Impose the kernel never to be negative. If an interpolated value is <0 we set them to be == 0.
- Solving conflicts w/r to JB version recently merged
- Adding some test coverage
- Improving docstrings a bit

### DIFF
--- a/linetools/spectra/lsf.py
+++ b/linetools/spectra/lsf.py
@@ -309,7 +309,7 @@ class LSF(object):
         # return lsf Table()
         return lsf
 
-    def interpolate_to_wv_array(self,wv_array, kind='Akima', debug=False):
+    def interpolate_to_wv_array(self, wv_array, kind='Akima', debug=False):
         """ Interpolate an LSF to a wavelength array.
         
         Given `wv_array` this function interpolates an LSF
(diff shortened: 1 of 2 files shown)


linetools/spectra/lsf.py
```diff
@@ -211,25 +210,25 @@ def load_COS_data(self):
         return pixel_scale, data
 
     def interpolate_to_wv0(self, wv0):
-        """Retrieves a unique LSF valid at wavelength wv0
+        """Retrieves an unique LSF valid at wavelength wv0
```
jnburchett: I think 'a unique' was correct?

linetools/spectra/lsf.py
```diff
@@ -364,17 +364,25 @@ def interpolate_to_wv_array(self,wv_array, kind='Akima'):
...
+        # make sure the kernel is never negative
+        cond = lsf_vals < 0
+        if np.sum(cond) > 0:
+            warnings.warn('The interpolated kernel has negative values; imposing them to be 0.')
+            if debug:
+                import matplotlib.pyplot as plt
+                plt.plot(wv_array_AA, lsf_vals, 'k-')
+                # import pdb; pdb.set_trace()
+            lsf_vals = np.where(lsf_vals < 0, 0., lsf_vals)
+
```
jnburchett: I plotted up a kernel generated with interpolate_to_wv_array, and it seems to me that the wings approach zero too steeply (which likely caused the original function to go negative).  Did you check an interpolated version against the original LSF from STScI to ensure they line up?

ntejos: I did check and the resulting interpolation is good as expected. Of course, the quality of the recovered LSF kernel depends on the input `wv_array`, if it is too coarse the output won't be nice. I have therefore added an extra check to make sure the LSF kernel is sampled by at least 10 pixels within `wv_array` (see lines 355-360 above).

PR #282, opened 2016-08-25
Mask the continuum array properly

lwymarie:


profxj:
Please provide a description, ie. why you have made this PR, what it does,
what else you added.  It can be brief but not zero.

profxj:
I've looked over the proposed changes and I prefer the way it was previously.  
The point is that the user should be matching the array that he/she gets from XSpectrum1D
without knowing whether that spectrum is padded with zeros etc.

You are, however, welcome to raise an error if someone inputs an array to
`value` that has s different size then the masked one.

PR #287, opened 2016-09-07
NANs in xspectrum1d.rebin

profxj:
Handles NANs in flux array

Test too


linetools/spectra/tests/test_xspectrum_utils.py
```diff
@@ -86,6 +86,10 @@ def test_rebin(spec):
     """
     imn = np.argmin(np.abs(newspec.wavelength-8055*u.AA))
     np.testing.assert_allclose(newspec.sig[imn].value, 0.0169634, rtol=1e-5)
+    # With NANs
+    spec.flux[100:110] = np.nan
```
ntejos: I do not think this works the way you are expecting (do not think you can reassign values this way).

ntejos:
Looks good apart from the inline comment regarding the test.

profxj:
Right you were.  Fixed.

profxj:
Merging.

Issue #285, opened 2016-08-30
ivarsmooth


jhennawi:
Can I request that we implement the IDL routine ivarsmooth as a method?

profxj:
Addressed.

PR #288, opened 2016-09-07
ivarsmooth method

profxj:
Introduce an inverse variance smoothing method.

Port of ivarsmooth from IDL

Includes test and sparse doc.

Addresses Issue #285


jhennawi:
Sounds good.

profxj:
Merging

PR #291, opened 2016-09-11
Header Extensions

profxj:
Allow one to grab the header from a different extension

Test too


linetools/spectra/io.py
```diff
@@ -25,7 +25,8 @@
 
 
 def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
-             format='ascii', exten=None, debug=False, select=0, **kwargs):
+             format='ascii', exten=None, debug=False, select=0,
+             head_ext=0, **kwargs):
```
ntejos: Could you please move the `head_ext` keyword right after `exten` keyword. Consider refactor `exten` -> `ext` (?)

profxj:
Moved and changed to head_exten

linetools/spectra/io.py
```diff
@@ -199,7 +202,11 @@ def readspec(specfil, inflg=None, efil=None, verbose=False, multi_ivar=False,
 
     # Add in the header
     #xspec1d.head = head0
-    xspec1d.meta['headers'][0] = head0
+    if head_ext == 0:
```
ntejos: I would have preferred to do these checks right before the line where `head0` is defined...

profxj: well, head0 is useful for a few things on its own

ntejos: Got it; then I suggest to store `head0` and e.g. `head_spec` variables both in the same part of the code. Please also add inline comments pointing out that `head0` contains info that `head_spec` may not necessarily have.

profxj: I don't know what `head_spec` is

ntejos: `head_spec` would be your current `head = hdulist[head_ext].header` (apologies for being pedantic...)

profxj:
Merging

PR #292, opened 2016-09-16
Contiuum edits

profxj:
Two changes related to continua:

1.  Enable reading of continuum spectrum of different size than data

2.  Modify lt_continuum to not mask by default (optional)

Test and files


linetools/isgm/utils.py
```diff
@@ -300,7 +300,7 @@ def iontable_from_components(components, ztbl=None, NHI_obj=None):
         mt = np.where((iontbl['Z'] == 1) & (iontbl['ion']==1))[0]
         if len(mt) == 1:
             iontbl[mt[0]]['logN'] = NHI_obj.NHI
-            iontbl[mt[0]]['sig_logN'] = NHI_obj.sig_NHI
+            iontbl[mt[0]]['sig_logN'] = np.mean(NHI_obj.sig_NHI) # Allow for two values
```
lwymarie: Why not keep both of the asymmetric sigma values?

profxj: It's only for this ion_N table and NHI tends to be the only one.

profxj:
Merging.

PR #296, opened 2016-10-07
radec_script

profxj:
Introduces lt_radec

a script to make it easy to give a coordinate and
have printed that corrdinate in several formats.

Includes docs and test

### DIFF
--- a/linetools/scripts/lt_radec.py
+++ b/linetools/scripts/lt_radec.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+
+"""
+Ingest input coordinates and print them to the screen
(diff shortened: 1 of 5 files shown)


linetools/scripts/lt_radec.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+
+"""
+Ingest input coordinates and print them to the screen
+Plot the data for a line (or lines)
+Also print the line data
```
jnburchett: What do the above two lines have to do with this script?  Is line data called somehow and I'm missing it?

Issue #300, opened 2016-10-19
readspec: issue with parse_linetools_spectrum_format()


ntejos:
There is an issue with `parse_linetools_spectrum_format()` when trying to read a fits table with continuum (generated by lt_continuumfit)::

> lt_xspec 3C273_cont.fits

```

/home/ntejos/anaconda/lib/python2.7/site-packages/matplotlib/__init__.py:872: UserWarning:  axes.color_cycle is deprecated and replaced with axes.prop_cycle; please use the latter.
   warnings.warn(self.msg_depr % (key, alt_key))
Traceback (most recent call last):
  File "/home/ntejos/anaconda/bin/lt_xspec", line 9, in <module>
    load_entry_point('linetools', 'console_scripts', 'lt_xspec')()
  File "/home/ntejos/python/linetools/linetools/scripts/lt_xspec.py", line 77, in main
    rsp_kwargs=rsp_kwargs, air=pargs.air)
  File "/home/ntejos/python/linetools/linetools/guis/xspecgui.py", line 60, in __init__
    rsp_kwargs=rsp_kwargs, **kwargs)
  File "/home/ntejos/python/linetools/linetools/guis/spec_widgets.py", line 71, in __init__
    rsp_kwargs=rsp_kwargs)
  File "/home/ntejos/python/linetools/linetools/guis/utils.py", line 207, in read_spec
    spec = lsx.XSpectrum1D.from_file(spec_fil, exten=exten, **kwargs['rsp_kwargs'])
  File "/home/ntejos/python/linetools/linetools/spectra/xspectrum1d.py", line 67, in from_file
    slf = tio.readspec(ifile, **kwargs)
  File "/home/ntejos/python/linetools/linetools/spectra/io.py", line 127, in readspec
    xspec1d = parse_linetools_spectrum_format(hdulist)
  File "/home/ntejos/python/linetools/linetools/spectra/io.py", line 617, in parse_linetools_spectrum_format
    xspec1d.meta.update(json.loads(hdulist[0].header['METADATA']))
  File "/home/ntejos/anaconda/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/home/ntejos/anaconda/lib/python2.7/json/decoder.py", line 367, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 1 column 10798 - line 1 column 10801 (char 10797 - 10800)
```

[...]

PR #301, opened 2016-10-19
Io patch

ntejos:
Ugly patch to read metadata from contpoints... This is a temporary solution that should be properly addressed; related to issue #300 .


profxj:
Ugly but ok.

Merging.

PR #312, opened 2016-11-02
Fix doc test

profxj:
Fixed by removing most of the tests in the doc string.

Not sure why these were ever passing..


profxj:
Merging.

PR #313, opened 2016-11-03
Xspectrum1d local s2n

ntejos:
Adds a method to calculate the average local signal-to-noise, with tests.

### DIFF
--- a/linetools/spectra/tests/test_xspec_methods.py
+++ b/linetools/spectra/tests/test_xspec_methods.py
@@ -45,17 +45,22 @@ def test_addmask():
     assert np.sum(spec.data['flux'][0].mask) > 3000
 
 
-# def test_get_local_s2n():
-if 1:
+def test_get_local_s2n():
     spec = XSpectrum1D.from_file(data_path('UM184_nF.fits'))
     wv0 = 4000 * u.AA
     s2n, sig_s2n = spec.get_local_s2n(wv0, 20, flux_th=0.9)
-    np.testing.assert_allclose(s2n, 6.2053003)
-    np.testing.assert_allclose(sig_s2n, 3.2064826)
+    np.testing.assert_allclose(s2n, 9.30119800567627)
+    np.testing.assert_allclose(sig_s2n, 1.0349911451339722)
     # test with continuum
-    smooth_spec = spec.gauss_smooth(len(spec.wa)/20.)
-
+    spec.co = np.ones_like(spec.flux)
+    s2n, sig_s2n = spec.get_local_s2n(wv0, 20, flux_th=0.9)
+    np.testing.assert_allclose(s2n, 10.330545425415039)
+    np.testing.assert_allclose(sig_s2n, 0.4250050187110901)
     # test errors
-    wv0 = 1200 * u.AA
+    # out of range
     with pytest.raises(ValueError):
         spec.get_local_s2n(1215*u.AA, 20)
+    # sig not defined
+    spec = XSpectrum1D.from_tuple((spec.wavelength, spec.flux))
+    with pytest.raises(ValueError):
+        spec.get_local_s2n(wv0, 20)
--- a/linetools/spectra/tests/test_xspec_methods.py
+++ b/linetools/spectra/tests/test_xspec_methods.py
@@ -49,16 +49,16 @@ def test_get_local_s2n():
     spec = XSpectrum1D.from_file(data_path('UM184_nF.fits'))
     wv0 = 4000 * u.AA
     s2n, sig_s2n = spec.get_local_s2n(wv0, 20, flux_th=0.9)
-    np.testing.assert_allclose(s2n, 9.30119800567627)
-    np.testing.assert_allclose(sig_s2n, 1.0349911451339722)
+    np.testing.assert_allclose(s2n, 9.30119800567627, rtol=1e-5)
+    np.testing.assert_allclose(sig_s2n, 1.0349911451339722, rtol=1e-5)
     # test with continuum
     spec.co = np.ones_like(spec.flux)
     s2n, sig_s2n = spec.get_local_s2n(wv0, 20, flux_th=0.9)
-    np.testing.assert_allclose(s2n, 10.330545425415039)
-    np.testing.assert_allclose(sig_s2n, 0.4250050187110901)
+    np.testing.assert_allclose(s2n, 10.330545425415039, rtol=1e-5)
+    np.testing.assert_allclose(sig_s2n, 0.4250050187110901, rtol=1e-5)
     # test errors
     # out of range
-    with pytest.raises(ValueError):
+    with pytest.raises(IOError):
         spec.get_local_s2n(1215*u.AA, 20)
     # sig not defined
     spec = XSpectrum1D.from_tuple((spec.wavelength, spec.flux))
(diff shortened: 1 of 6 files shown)


linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+    def get_local_s2n(self, wv0, npix=51, flux_th=0., debug=False):
+        """It computes the local average signal-to-noise (s2n) over npix pixels around wv0.
+        If `flux_th` is given, pixels with fluxes below spec.flux*flux_th are masked out.
+
+        Parameters
+        ----------
+        wv0 : Quantity
+            Observed wavelength where to perform the calculation.
+        npix : float or int, optional
+            Number of pixels to perform the calculation (does not
```
profxj: Not sure what 'does not take into account...' means.  Does it mean you always assess 51 pixels or you will assess less if some are masked?

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+        -------
+        s2n : float
+            Local average signal-to-noise
+        s2n_sig : float
+            Standard deviation of the s2n measurement
+        """
+
+        # Do some checks
+        if (wv0 < self.wvmin) or (wv0 > self.wvmax):
+            raise ValueError("`wv0` is outside spectral range.")
```
profxj: IOError

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+        if not self.sig_is_set:
+            raise ValueError("Spectrum has not defined an error array; cannot compute signal-to-noise.")
+        npix = int(npix)
```
profxj: The doc says this will be an odd integer.  That isn't the case here.

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+
+        # renaming
+        wa = self.wavelength
```
profxj: why rename?  This makes the code harder to read.

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+        fl = self.flux
+        er = self.sig
+
+        # if flux_th is float, then check whether continuum exists
+        # otherwise we estimate it.
+        if isinstance(flux_th, float):
+            if self.co_is_set:
+                flux_th = self.co * flux_th
```
profxj: use a different variable name instead of recasting flux_th.  Best to preserve what was input.

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+            else:
```
profxj: Add a comment here as to what you are doing.  Looks like you are generating a local continuum.

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+        elif len(flux_th) != len(self.flux):
+            raise ValueError('`flux_th` must be either float or array of same shape as self.flux.')
+
+        #find pixel index for wv0
+        cond = np.fabs(wv0 - wa) == np.min(np.fabs(wv0 - wa))
+        ind = np.where(cond)[0][0]
+
+        # define the wa window with at least npix pixels
+        i = 0
+        while True:
```
profxj: while loops scare me (one can get caught forever) and this code is pretty hard to read (and not Pythonic).  try something like this:

```
gd_flux = self.flux > flux_th
if np.sum(gdp) < npix:
   raise ValueError("No good")
gd_idx = np.where(gd_flux)[0]
idx_diff = np.abs(gd_idx - ind)
# Now sort on idx_diff and grab the closest npix indices and use those
```

profxj:
Added a bunch of inline comments.  Otherwise looks good.
Please add the method to the xspectrum1d docs too.

Am nervous about the astropy_helpers stuff but hopefully it is ok.

linetools/spectra/tests/test_xspec_methods.py
```diff
@@ -44,3 +44,29 @@ def test_addmask():
     spec.add_to_mask(badp, compressed=True)
     assert np.sum(spec.data['flux'][0].mask) > 3000
 
+
+def test_get_local_s2n():
+    spec = XSpectrum1D.from_file(data_path('UM184_nF.fits'))
+    wv0 = 4000 * u.AA
+    s2n, sig_s2n = spec.get_local_s2n(wv0, 20, flux_th=0.9)
+    np.testing.assert_allclose(s2n, 9.30119800567627)
```
ntejos: Done.

linetools/spectra/xspectrum1d.py
ntejos: Always assess 51 pixels (when possible), so the range is increased if there are masked pixels.

linetools/spectra/xspectrum1d.py
ntejos: Done.

linetools/spectra/xspectrum1d.py
ntejos: Docstring is updated.

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+        if not self.sig_is_set:
+            raise ValueError("Spectrum has not defined an error array; cannot compute signal-to-noise.")
+        npix = int(npix)
+
+        # renaming
+        wa = self.wavelength
+        fl = self.flux
+        er = self.sig
+
+        # if flux_th is float, then check whether continuum exists
```
ntejos: Do not follow this comment...

linetools/spectra/xspectrum1d.py
ntejos: Done.

linetools/spectra/xspectrum1d.py
ntejos: Done. Indeed, we just smooth the original spectrum to have a rough continuum estimation.

linetools/spectra/xspectrum1d.py
```diff
@@ -963,6 +963,97 @@ def stitch(self, idx=None, scale=1.):
...
+                n_smooth = np.max([10*npix, 500])
+                smooth_spec = self.gauss_smooth(n_smooth)
+                co = smooth_spec.flux
+                flux_th = co * flux_th
+        # if it is not float, it should be array like self.flux
+        elif len(flux_th) != len(self.flux):
+            raise ValueError('`flux_th` must be either float or array of same shape as self.flux.')
+
+        #find pixel index for wv0
+        cond = np.fabs(wv0 - wa) == np.min(np.fabs(wv0 - wa))
```
ntejos: Looks good, thanks.

linetools/spectra/xspectrum1d.py
profxj: Rewrite the code so that flux_th = None initially instead of 0.

linetools/spectra/xspectrum1d.py
ntejos: Very nice, thanks.

profxj:
Looks great.  Not sure why coverage went down, but merge when the tests pass.

PR #322, opened 2016-11-22
dict in meta for XSpectrum1D

profxj:
Allows for using dict objects instead of Header objects
in XSpectrum1D meta.

Should probably add a test..


profxj:
merging

PR #324, opened 2016-11-22
Analysis abslines

ntejos:
- Added function to calculate tau0 at line center.
- Added function to estimate equivalent width from logN and b for a given transition.
- Both with tests

Please review but do not merge yet; I intend to add a couple of extra ones that build from these.

### DIFF
--- a/linetools/analysis/absline.py
+++ b/linetools/analysis/absline.py
@@ -301,16 +301,27 @@ def get_tau0(wa0, fosc, logN, b):
         Rest-frame wavelength of the transition
     fosc : float
         Oscillator strength of the transition
-    logN : float
+    logN : float, or np.array
         log10 of the column density in cm^{-2}
-    b : Quantity
+    b : Quantity, or Quantity array (same shape as logN)
         Doppler parameter
 
     Returns
     -------
-    tau0: float
-        Optical depth at the line center
+    tau0: float or array
+        Optical depth at the line center. If logN and b are
+        arrays they must be of same shape.
     """
+    # check format for logN and b
+    try:
+        n = len(logN)
+        if n != len(b):
+            raise IOError('If logN is array, b must be array of same shape.')
+    except TypeError:  # assuming float for logN
+        if not isinstance(logN, (float, int)):
+            raise IOError('If logN is not array, it must be float.')
+    # force to numpy array
+    logN = np.array(logN)
 
     # convert units to CGS
     b_cgs = b.to('cm/s').value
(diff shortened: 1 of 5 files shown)


linetools/analysis/absline.py
```diff
@@ -284,3 +287,101 @@ def sum_logN(obj1, obj2):
...
+    emission which is fine for IGM or ISM except for radio-frequency
+    transitions.
+
+    Parameters
+    ----------
+    wa0 : Quantity
+        Rest-frame wavelength of the transition
+    fosc : float
+        Oscillator strength of the transition
+    logN : float, or np.array
```
profxj: Should we read in as N instead of logN to keep units sensible?

linetools/analysis/absline.py
```diff
@@ -284,3 +287,101 @@ def sum_logN(obj1, obj2):
...
+        Doppler parameter
+
+    Returns
+    -------
+    tau0: float or array
+        Optical depth at the line center. If logN and b are
+        arrays they must be of same shape.
+    """
+    # check format for logN and b
+    try:
```
profxj: Could use astropy's  `isiterable` although this is fine too.

linetools/analysis/absline.py
```diff
@@ -284,3 +287,101 @@ def sum_logN(obj1, obj2):
...
+        n = len(logN)
+        if n != len(b):
+            raise IOError('If logN is array, b must be array of same shape.')
```
profxj: You aren't checking the shape, only the length.

linetools/analysis/absline.py
```diff
@@ -284,3 +287,101 @@ def sum_logN(obj1, obj2):
...
+    # force to numpy array
+    logN = np.array(logN)
+
+    # convert units to CGS
+    b_cgs = b.to('cm/s').value
+    wa0_cgs = wa0.to('cm').value
+    N_cgs = 10**logN  # assumed to be in cm^2
+
+    # tau0
+    tau0 = np.sqrt(np.pi) * e2_me_c_cgs * N_cgs * fosc * wa0_cgs  / b_cgs  # eq. 9.8 Draine 2011
```
profxj: I'd use units the way they prefer to be used.   That is, keep them all (instead of taking values) and do a `decompose()` on tau0 to insure it is unitless.  This means adding cm^-2 on N_cgs (or better, having it read in with units)

linetools/analysis/absline.py
```diff
@@ -284,3 +287,101 @@ def sum_logN(obj1, obj2):
...
+    """ For a given transition with fosc and gamma, it
+    returns the rest-frame equivalent width for a given
+    logN and b. It uses the approximation given by Draine 2011 book
+    (eq. 9.27), which comes from atomic physics considerations
+    See also Rodgers & Williams 1974 (NT: could not find the reference
+    given by Draine)
+
+    Parameters
+    ----------
+    logN : float
```
profxj: Again, we might require using N here (with units)

linetools/analysis/absline.py
```diff
@@ -284,3 +287,101 @@ def sum_logN(obj1, obj2):
...
+    W_satu = np.sqrt(W2_satu)  # dimensionless
+    W = np.where(tau0 <= 1.25393, W_thin, W_satu)
+
+    # if tau0 <= 1.25393:
+    #     W = np.sqrt(np.pi) * (b_cgs/c_cgs) * tau0 / (1 + tau0 / (2 * np.sqrt(2))) # dimensionless
+    # else:
+    #     W2 = (2 * b_cgs/c_cgs)**2 * np.log(tau0 / np.log(2)) \
+    #         + (b_cgs/c_cgs) * (wa0_cgs * gamma_cgs / c_cgs) * (tau0 - 1.25393) / np.sqrt(np.pi)  # dimenonless
+    #     W = np.sqrt(W2)  # dimensionless
+    Wr = W * wa0  # in wavelength units (see equation 9.4 of Draine)
```
profxj: again, I encourage you to use units here instead of stripping down to values.   Otherwise, you are somewhat defeating the point of using units.

linetools/analysis/absline.py
ntejos: Good suggestion.

linetools/analysis/absline.py
ntejos: Thanks.

linetools/analysis/absline.py
ntejos: You are right, updating code.

linetools/analysis/absline.py
ntejos: Good idea.

linetools/analysis/absline.py
ntejos: Sure.

linetools/analysis/absline.py
ntejos: Good point.

ntejos:
@profxj are you happy with the changes?

linetools/analysis/absline.py
```diff
@@ -284,3 +290,155 @@ def sum_logN(obj1, obj2):
...
+        Optical depth at the line center. If logN and b are
+        arrays they must be of same shape.
+    """
+    # check format for N and b
+    if isiterable(N):
+        if np.shape(N) != np.shape(b):
+            raise IOError('If N is array, b must be array of same shape.')
+
+    # convert to CGS
+    b_cgs = b.to('cm/s')
```
profxj: These lines aren't necessary.  astropy units will do it for you.

profxj:
Added one additional comment on units.

I was also assuming you'd put in a few wrapper methods within AbsLine 
to call these from within the class.  Can you?

Am approving anyhow.

linetools/analysis/absline.py
ntejos: I'd like to leave these lines though because they help to check for the correct units for the input parameters, and will raise meaningful errors for the user if the units are wrong.

ntejos:
What kind of methods do you have in mind?

profxj:
Nothing too exciting.

something like:

aline.get_tau0()

ntejos:
Sounds good, will add it now.

profxj:
And how about

aline.get_Wr_from_N_b()  ?

With that, merge. :)

ntejos:
Sure thing.

Issue #307, opened 2016-10-27
Tie attrib['z'] to LineLimits._z


profxj:
Hopefully this can be done.

ntejos:
I think we can close this issue, right?

profxj:
Addressed by PR #326

PR #326, opened 2016-11-24
Refactor z in SpectralLine

profxj:
Attemps to expunge usage of attrib['z'] in SpectralLine

Wound up refactoring LineLimits too, a bit.

Lots of edtis to docs and tests.

Addresses Issue #307

### DIFF
--- a/docs/specline.rst
+++ b/docs/specline.rst
@@ -35,17 +35,20 @@ The base attributes for the SpectralLine class are:
 ================ ================= ========== ========================================
 Property         Variable          Type       Description
 ================ ================= ========== ========================================
+Limits           limits            LineLimits The redshift and limits of the line in redshift, velocity
+                                              (w/r to its redshift) and observed wavelength.
 RA, Dec          attrib['coord']   Coord      astropy.coordinate
-Redshift         attrib['z']       float      Reference redshift
-Redshift sigma   attrib['sig_z']   float      Reference redshift uncertainty
 Velocity         attrib['v']       Quantity   line velocity relative to its redshift
 Velocity sigma   attrib['sig_v']   Quantity   1 sigma uncertainty in the velocity
 Equivalent Width attrib['EW']      Quantity   Equivalent width
 EW sigma         attrib['sig_EW']  Quantity   1 sigma uncertainty in EW
 EW flag          attrib['flag_EW'] int        Equivalent width flag
-Limits           limits            LineLimits The limits of the line in redshift, velocity
-                                              (w/r to its redshift) and observed wavelength.
 ================ ================= ========== ========================================
+
+Note that redshift is sufficiently important that it is contained
+within its own object.  It is also accessible as a property, e.g.::
+
+   z = sline.z
 
 .. _specanalysis
 
--- a/linetools/analysis/linelimits.py
+++ b/linetools/analysis/linelimits.py
@@ -28,6 +28,7 @@ class LineLimits(object):
     vlim : Quantity array
       velocity limits for the line in km/s
     """
+
     @classmethod
     def from_specline(cls, aline, z, zlim):
         """ From AbsLine or Emline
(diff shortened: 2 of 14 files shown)


docs/specline.rst
```diff
@@ -35,18 +35,21 @@ The base attributes for the SpectralLine class are:
 ================ ================= ========== ========================================
 Property         Variable          Type       Description
 ================ ================= ========== ========================================
+Limits           limits            LineLimits The redshift and limits of the line in redshift, velocity
+                                              (w/r to its redshift) and observed wavelength.
 RA, Dec          attrib['coord']   Coord      astropy.coordinate
-Redshift         attrib['z']       float      Reference redshift
-Redshift sigma   attrib['sig_z']   float      Reference redshift uncertainty
```
ntejos: Would it be worth to have a attrib['z_obs'] and attrib['sig_z_obs'] to allow the user to store a 'measured' redshift within the line? Not particularly strong about this though.

linetools/analysis/linelimits.py
```diff
@@ -25,8 +28,9 @@ class LineLimits(object):
     vlim : Quantity array
       velocity limits for the line in km/s
```
ntejos: Remove 'in km/s' and add something like "at the rest-frame of z".

linetools/analysis/linelimits.py
```diff
@@ -68,29 +90,41 @@ def __init__(self, wrest, z, zlim, **kwargs):
         self.set(zlim, **kwargs)
```
ntejos: delete "Should match line.attrib['z']" in docstring above.

linetools/analysis/linelimits.py
```diff
@@ -25,8 +28,9 @@ class LineLimits(object):
     vlim : Quantity array
       velocity limits for the line in km/s
     """
+
     @classmethod
-    def from_specline(cls, aline, zlim):
+    def from_specline(cls, aline, z, zlim):
```
ntejos: But now we require the absline to always have limits defined and redshift defined. Couldn't we set the limits 'from_specline' without the need to input `z`? i.e. 'z' will be read directly from `aline.z`

linetools/spectralline.py
```diff
@@ -151,18 +150,15 @@ def from_dict(cls, idict, coord=None, warn_only=False, chk_data=True, **kwargs):
...
-                    qlim = ltu.convert_quantity_in_dict(idict['limits'][key])
-                    sline.limits.set(qlim)
-                    break  # only one limit is needed to define them all
-                else:
-                    sline.limits.set(idict['limits'][key])  # this is zlim
-                    break  # only one limit is needed to define them all
-        except KeyError:
+        # Set z and limits
+        if 'z' in sline.attrib.keys():
+            z = sline.attrib.pop('z')
```
ntejos: Is this for backwards compatibility, right? If so please add inline comment.

linetools/spectralline.py
```diff
@@ -425,8 +425,8 @@ def measure_restew(self, **kwargs):
         self.measure_ew(**kwargs)
 
         # Push to rest-frame
-        self.attrib['EW'] = self.attrib['EW'] / (self.attrib['z']+1)
-        self.attrib['sig_EW'] = self.attrib['sig_EW'] / (self.attrib['z']+1)
+        self.attrib['EW'] = self.attrib['EW'] / (self.z+1)
+        self.attrib['sig_EW'] = self.attrib['sig_EW'] / (self.z+1)
```
ntejos: Now that I think about it, shouldn't we have 'observed EW' and 'rest-frame EW' as separate attributes?

linetools/spectralline.py
```diff
@@ -266,7 +267,6 @@ def setz(self, z):
         if not isinstance(z,float):
             raise IOError("Input redshift needs to be a float")
         # Set
-        self.attrib['z'] = z
         self.limits._z = z
         # Warning?
         if self.limits.is_set():
```
ntejos: The update should be done automatically now, right?

docs/specline.rst
profxj: I expect this will happen and should, but I decided I'd rather
not make it a default attribute.

As for name, z_measured might be best.

linetools/analysis/linelimits.py
profxj: No.  aline.z is a pointer to LineLimits._z
That is the point here, i.e. we want one redshift value set and used.

It does require these to be defined, but they are defaulted to 0 and (0,0)

linetools/analysis/linelimits.py
profxj: agreed

linetools/spectralline.py
profxj: added

linetools/spectralline.py
profxj: probably, but not in this PR.  Feel free to raise an Issue

linetools/spectralline.py
profxj: This allows you to modify the redshift of the line
without understanding how to use LineLimit

ntejos:
Please merge whenever you are happy.

profxj:
Merging

PR #329, opened 2016-11-24
Voigt: Voigt model from component list

ntejos:
As stated.

### DIFF
--- a/linetools/analysis/voigt.py
+++ b/linetools/analysis/voigt.py
@@ -251,7 +251,7 @@ def voigt_from_abslines(iwave, line, fwhm=None, ret=['vmodel'],
     return ret_val
 
 
-def voigt_from_components(wv_array, complist, fwhm=None):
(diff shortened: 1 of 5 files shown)


linetools/analysis/voigt.py
```diff
@@ -250,6 +251,43 @@ def voigt_from_abslines(iwave, line, fwhm=None, ret=['vmodel'],
...
+    wvmin = np.min(wv_array)
+    wvmax = np.max(wv_array)
+    gdlin = []
+    for comp in complist:
+        for line in comp._abslines:
+            wvobs = (1 + line.z) * line.wrest
+            if (wvobs > wvmin) & (wvobs < wvmax):
+                gdlin.append(line)
+    # return the model
+    return voigt_from_abslines(wv_array, gdlin, fwhm=fwhm, ret=['vmodel'])
```
profxj: I suggest you add **kwargs here and in the method call.

Issue #317, opened 2016-11-10
LineList: implement rel_strenght and all transitions for molecules


ntejos:
- Implement relative strength criteria for molecules.
- Implement all_transitions method for molecules.

ntejos:
Addressed in PR 332.

PR #332, opened 2016-12-02
Llist h2: implement basic rel_strength and all_transition() method

ntejos:
Addresses Issue #317 

Working on this, I came to the conclusion that we should revisit the way things are done internally within LineList to treat atomic transitions versus molecular transitions. We may need redefine better namings for columns in self._data. Probably best to leave for a separate issue.

### DIFF
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -331,19 +331,19 @@ class LineList(object):
         for sorting the underlying data table in convenient ways, e.g. by expected
         relative strength, abundance, etc.
 
-        * For atomic transitions these new column include:
+        * For atomic transitions these new columns include:
             - `ion_name` : HI, CIII, CIV, etc
             - `log(w*f)` : np.log10(wrest * fosc)  # in np.log10(AA)
             - `abundance` : either [`none`, `solar`]
             - `ion_correction` : [`none`]
             - `rel_strength` : log(w*f) + abundance + ion_correction
 
-        * For molecules a different approach is done:
+        * For molecules a different approach is used:
             - `ion_name` : B0-0P, C6-0, etc.
             - `rel_strength`: We have only three arbitrary levels: [1, 50, 100]
-                    100 is for Jj={0,1} and Jk={0,1}
-                    50 is for Jj={2,3} and Jk={2,3}
-                    1 is for the rest
+                    100 is for Jk={0,1}
+                    50 is for Jk={2,3}
+                    1 otherwise
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


linetools/lists/linelist.py
```diff
@@ -329,14 +329,22 @@ def set_extra_columns_to_datatable(self, abundance_type='solar', ion_correction=
                                        redo=False):
         """Sets new convenient columns to the self._data QTable. These will be useful
         for sorting the underlying data table in convenient ways, e.g. by expected
-        relative strength.
-        These new column include:
+        relative strength, abundance, etc.
+
+        * For atomic transitions these new column include:
```
profxj: new columns

linetools/lists/linelist.py
```diff
@@ -329,14 +329,22 @@ def set_extra_columns_to_datatable(self, abundance_type='solar', ion_correction=
...
             - `ion_name` : HI, CIII, CIV, etc
             - `log(w*f)` : np.log10(wrest * fosc)  # in np.log10(AA)
             - `abundance` : either [`none`, `solar`]
             - `ion_correction` : [`none`]
             - `rel_strength` : log(w*f) + abundance + ion_correction
 
+        * For molecules a different approach is done:
```
profxj: is used

linetools/lists/linelist.py
```diff
@@ -329,14 +329,22 @@ def set_extra_columns_to_datatable(self, abundance_type='solar', ion_correction=
...
+            - `ion_name` : B0-0P, C6-0, etc.
+            - `rel_strength`: We have only three arbitrary levels: [1, 50, 100]
+                    100 is for Jj={0,1} and Jk={0,1}
```
profxj: Reviewing H2 a bit, I see that the key is really Jk.  I'd do this:

100 for Jk = 0,1
50 for Jk = 2,3
1 otherwise

linetools/lists/linelist.py
```diff
@@ -349,25 +357,40 @@ def set_extra_columns_to_datatable(self, abundance_type='solar', ion_correction=
...
         # Set ion_name column
-        ion_name = [name.split(' ')[0] for name in self.name]
+        if self.list in ['HI', 'ISM', 'EUV', 'Strong']:
+            ion_name = [name.split(' ')[0] for name in self.name]  # valid for atomic transitions
+        elif self.list in ['H2']:
+            ion_name = [name.split('(')[0] for name in self.name]  # valid for H2
         self._data['ion_name'] = ion_name
 
+        if self.list in ['H2']:
+            # we want Jj and Jk to be 1 or 0 first
```
profxj: see above on docs.  I think we should just query on Jk

ntejos:
Thanksvery much! Indeed the new recovered ordered list of transitions is better now when I looked at J1018 in igmguesses panels. Please merge if your are happy with this.

PR #333, opened 2016-12-02
Abscomponents: complist_from_table and other

ntejos:
- complist_from_table() 
- get_components_at_z()

These two simple general methods added, with tests.

### DIFF
--- a/linetools/isgm/utils.py
+++ b/linetools/isgm/utils.py
@@ -23,7 +23,7 @@ from linetools.analysis import absline as ltaa
 from linetools.isgm.abscomponent import AbsComponent
 from linetools.spectralline import init_analy
 from linetools.abund.ions import name_ion
-
+from linetools import utils as ltu
 
 def chk_components(components, chk_match=False, chk_A_none=False, tol=0.2*u.arcsec):
     """ Performs checks on a list of components
--- a/linetools/isgm/utils.py
+++ b/linetools/isgm/utils.py
@@ -17,6 +17,7 @@ from astropy import constants as const
 from astropy import units as u
 from astropy.table import Table
 from astropy.units import Quantity
+from astropy.coordinates import SkyCoord
 
 from linetools.analysis import absline as ltaa
 from linetools.isgm.abscomponent import AbsComponent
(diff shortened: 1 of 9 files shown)


linetools/isgm/utils.py
```diff
@@ -239,6 +241,73 @@ def xhtbl_from_components(components, ztbl=None, NHI_obj=None):
     tbl = Table()
     #
 
+def complist_from_table(table):
+    """
+    Returns a list of AbsComponents from an input table.
```
profxj: table -> Table

linetools/isgm/utils.py
```diff
@@ -239,6 +241,73 @@ def xhtbl_from_components(components, ztbl=None, NHI_obj=None):
...
+        to a component)
+
+    Returns
+    -------
+    complist : list
+        List of AbsComponents defined from the input table.
+
+    Notes
+    -----
+    Mandatory column names: 'RA_deg', 'DEC_deg', 'ion_name', 'z_comp', 'vmin_kms', 'vmax_kms'
```
profxj: I have tended to not include units in the column names, i.e.
  RA instead of RA_deg
  vmin instead of vmin_kms

I'm not opposed to your approach, though.  This is more of a comment.

linetools/isgm/utils.py
```diff
@@ -376,6 +445,34 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+    """In a given list of components, it finds
+    the ones that are within dvlims from a given redshift
+    and returns a list of those components
+
+    Parameters
+    ----------
+    complist : list
+        List of AbsComponents
+    z : float
+        Redshift there to look for components
```
profxj: Redshift to search for components

linetools/isgm/utils.py
```diff
@@ -376,6 +445,34 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
+    dvlims : Quantity array
+        Rest-frame velocity limits around z
+        to look for components
+
+    Returns
+    -------
+    components_at_z : list
+        List of components within complist within dvlims from z
+    """
+    good_complist = []
```
profxj: Might do a little error checking here, especially on dvlims
to insure it is a Quantity array with velocity units.

linetools/isgm/utils.py
ntejos: I understand, but would prefer to make it explicit what units the function expects, otherwise I would need to do some extra checks that I would prefer to avoid :/ I thought using QTables and get the units directly from the column units, but I think it is somewhat less clear for the user.

profxj:
Give some thought to my one addition and/or merge.

ntejos:
Will add documentation soon for these before merging.

PR #335, opened 2016-12-09
XSpectrum1D slicing, and more

profxj:
Enables slicing of XSpectrum1D

Deals with co in rebinning
Halts rebinning if rejected pixels are included

Avoids dividing by co=0. values in normalized data

New docs, new tests


linetools/spectra/utils.py
```diff
@@ -292,11 +304,11 @@ def rebin(spec, new_wv, do_sig=False, all=False, **kwargs):
...
     # update continuum
-    if spec.co_is_set:
-        x, y = spec._get_contpoints()
-        new_co = spec._interp_continuum(x, y, new_wv)
-    else:
-        new_co = None
+    new_co = None
+    if do_co:
+        if spec.co_is_set:
+            x, y = spec._get_contpoints()
```
jnburchett: In my testing example, I instantiated an XSpectrum1D with a fresh (no continuum points in the header) FITS file, fit a continuum with fit_continuum(), and tried `spec._get_contpoints()`.  The resulting error message is appended below.  Even though `spec.co_is_set` is now `True`, there is no `spec.meta['contpoints']` (returns KeyError.

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-33-b2e0f623c450> in <module>()
----> 1 x,y = spec._get_contpoints()

/Users/burchett/Dropbox/Code/linetools/linetools/spectra/xspectrum1d.pyc in _get_contpoints(self)
   1484             contpoints = self.meta['contpoints']
   1485         except:
-> 1486             raise RuntimeError('Meta data `contpoints` does not exist; cannot get contpoints!')
   1487         xy = np.array(contpoints)
   1488         xy = xy.transpose()

RuntimeError: Meta data `contpoints` does not exist; cannot get contpoints!
```

linetools/spectra/utils.py
```diff
@@ -292,11 +304,11 @@ def rebin(spec, new_wv, do_sig=False, all=False, **kwargs):
...
+            new_co = spec._interp_continuum(x, y, new_wv)
```
jnburchett: Could we just rebin the continuum using the continuum (`spec.co`) itself instead of the knots, which we will no longer be storing in the header?

linetools/spectra/xspectrum1d.py
```diff
@@ -1421,11 +1426,13 @@ def fit_continuum(self, knots=None, edges=None, wlim=None, dw=10.,
 
         print('Updating continuum.')
         self.co = wrapper.continuum  # this should work with the assignment, even is self.co_is_set is False
-        if 'contpoints' not in self.meta:
-            self.meta['contpoints'] = []
-        self.meta['contpoints'].extend(
-            [tuple(pts) for pts in wrapper.contpoints])
-        self.meta['contpoints'].sort()
+        # Put into meta -- has been causing trouble
+        if False:
```
jnburchett: Should this line be `if self.co_is_set is False:'?

linetools/spectra/utils.py
profxj: So replaced.

linetools/spectra/xspectrum1d.py
profxj: No, I am now explicitly avoiding writing the continuum to the header.
Was causing too much grief.

ntejos: Fine by me.

PR #338, opened 2016-12-22
AbsComponent H2

ntejos:
A patch for AbsComponents to work (at a basic level without checks) with H2 lines. This PR is only  meant to make pyigm's IGMguesses to be able to handle H2 molecules. A more fundamental solution is beyond the scope of this PR.

### DIFF
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -198,7 +198,7 @@ class AbsComponent(object):
                 except:
                     raise IOError("Need to provide 'stars' parameter.")
             self.name = '{:s}_z{:0.5f}'.format(iname, self.zcomp)
-        elif (self.Zion != (-1, -1)):
+        elif (name is None) and (self.Zion == (-1, -1)):
             self.name = 'mol_z{:0.5f}'.format(self.zcomp)
         else:
             self.name = name
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -190,7 +190,7 @@ class AbsComponent(object):
             self.sig_logN = 0.
 
         # Name
-        if name is None:
+        if (name is None) and (self.Zion != (-1, -1)):
             iname = ions.ion_name(self.Zion, nspace=0)
             if self.Ej.value > 0:  # Need to put *'s in name
                 try:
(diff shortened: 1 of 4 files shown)


linetools/isgm/abscomponent.py
```diff
@@ -190,14 +190,16 @@ def __init__(self, radec, Zion, zcomp, vlim, Ej=0./u.cm, A=None,
             self.sig_logN = 0.
 
         # Name
-        if name is None:
+        if (name is None) and (self.Zion != (-1, -1)):
```
profxj: Can you add to the doc string that -1,-1 is now special?

linetools/isgm/utils.py
```diff
@@ -325,6 +326,55 @@ def complist_from_table(table):
     return complist
 
 
+def table_from_complist(complist):
+    """
+    Returns a QTable from an input list of AbsComponents. It only
```
profxj: I have to admit that I'm no longer too enamored with QTable's.  I was at the start but find that normal Table objects do 99% of what is needed.

Do we need one here?

linetools/isgm/utils.py
```diff
@@ -325,6 +326,55 @@ def complist_from_table(table):
...
+    fills in mandatory and special attributes (see notes below).
+    Information stored in dictionary AbsComp.attrib is ignored.
+
+    Parameters
+    ----------
+    complist : list of AbsCOmponents
```
profxj: AbsComponents

linetools/isgm/abscomponent.py
ntejos: Sure thing.

linetools/isgm/utils.py
ntejos: You are right; I find particularly annoying that vstack does not work properly with QTables. Will use Tables as default then. Still, there will be one place where QTable is more useful (e.g. to access entry values as quantity objects), but will be handled internally in the corresponding function; the user will always get a Table object out.

linetools/isgm/utils.py
ntejos: thanks.

profxj:
Failing Python 3 tests.  Fix and then merge.

PR #340, opened 2016-12-28
Small fixes to satisfy astropy v1.3

profxj:
As titled.  A few QTable manoeuvers for line lists.

No new tests or docs.


profxj:
Merging to pass tests on Travis with astropy v1.3

PR #341, opened 2016-12-29
Features for analysis.plots.stack_plot()

jnburchett:
Added options to return matplotlib Figure instance of stackplot and apply tight_layout

### DIFF
--- a/linetools/analysis/plots.py
+++ b/linetools/analysis/plots.py
@@ -36,6 +36,11 @@ def stack_plot(abslines, vlim=[-300,300.]*u.km/u.s, nrow=6, show=True,
       If True, return stackplot as plt.Figure() instance for further manipulation
     tight_layout : bool, optional
       If True, remove whitespace between panels
+
+    Returns
+    -------
+    fig : matplotlib Figure, optional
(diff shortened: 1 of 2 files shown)


linetools/analysis/plots.py
```diff
@@ -14,7 +14,7 @@
 import matplotlib as mpl
 
 def stack_plot(abslines, vlim=[-300,300.]*u.km/u.s, nrow=6, show=True,
-               ymnx=(-0.1,1.1), figsz=(18,11)):
+               ymnx=(-0.1,1.1), figsz=(18,11), return_fig=False, tight_layout=False):
```
profxj: Please add Returns docs.

linetools/isgm/abscomponent.py
```diff
@@ -713,14 +713,20 @@ def repr_joebvp(self, specfile, flags=(2,2,2), b_default=10*u.km/u.s):
             s += '\n'
         return s
 
-    def stack_plot(self, **kwargs):
+    def stack_plot(self, return_fig=False, **kwargs):
```
profxj: Also add Returns doc here.

PR #343, opened 2017-01-04
Order the components by z after reading from_dict

profxj:
As titled.

Deals with the arbitrary ordering of dict keys.  This could
create unusual (and at the least unpredictable) behavior
when reading comopnents from disk.

Added a new test file and test.


profxj:
Passed tests.  merging

PR #344, opened 2017-01-07
Collate in XSpectrum1D

profxj:
Updates collate() method.  Allows for multiple spectra
in each input XSpectrum1D object.

Updates tests and docs


profxj:
Merging

PR #345, opened 2017-01-12
laa: added N_from_Wr() and N_from_Wr_transition() with tests.

ntejos:
N_from_Wr() and N_from_Wr_transition() added to linetools.analysis.absline; these are useful for getting upper limits on N from measured Wr upper limits. Assume tau0<<1 where the relation between N and Wr is independent on Doppler parameter and gamma.

### DIFF
--- a/linetools/analysis/absline.py
+++ b/linetools/analysis/absline.py
@@ -440,6 +440,42 @@ def Wr_from_N(N, wrest, fosc):
     Wr = W * wrest  # in wavelength units (see equation 9.4 of Draine)
     return Wr
 
+def N_from_Wr(Wr, wrest, fosc):
+    """For a given transition with wrest and fosc, it
+    returns the column density N, for a given rest-frame equivalent width
+    Wr. This is an approximation only valid for tau0 << 1, where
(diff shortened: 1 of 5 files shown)


linetools/analysis/absline.py
```diff
@@ -531,3 +567,47 @@ def Wr_from_N_transition(N, transition, llist='ISM'):
     # return
     return Wr_from_N(N, wrest, fosc)
 
+
+def N_from_Wr_transition(Wr, transition, llist='ISM'):
```
profxj: While over-loading input statements is a bit dangerous, I might
prefer it here, i.e. combine these 2 methods and the 2nd input
can either be a Quantity (wrest) or a str (transition).  Could even
allow an AbsLine.

ntejos: I would rather not overload this the input statement as it goes somewhat against my preferred direction when coding. Instead, I have added the corresponding method to the AbsLine class, so you can use it directly from there. I hope you agree with this decision.

ntejos:
Have also added some basic documentation in AbsLine.rst regarding these methods.

profxj:
Looks good.  Please merge

PR #346, opened 2017-01-13
Add EW measurement in xspec

profxj:
Also adds a small QtWrapper for ExamineSpec
so that it may be launched by other GUI's (e.g. igmguesses).

No new docs or tests.


profxj:
Merging

PR #351, opened 2017-02-04
EW using do_analysis

profxj:
Require do_analysis=1 when analyzing EWs in AbsSystem

Also a bug fix, new docs and tests.


profxj:
Merging

PR #353, opened 2017-02-10
AbsSys widget

profxj:
Added a new widget to the spec_widgets module.

It is primarily to support line-fitting guis which will locate
in pyigm.

No docs or tests


profxj:
Merging.

PR #357, opened 2017-02-11
Scrub pip install for now

profxj:
As titled.

The pip wheel we have in PyPi is woefully out of date.
Until we figure out how to update (may require Neil) 
the install should be via git clone.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -44,15 +44,12 @@ installed with pip.
 Installing Linetools
 ====================
 
-If you plan to play around with the code and possibly contribute
-changes, then follow the instructions in the section below,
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -44,15 +44,12 @@ installed with pip.
...
+out of date.  We will try to update before long.  But for now
+follow the instructions in the section below,
+:ref:`installsource` to install linetools.
 
-    pip install linetools
-
-and you're done!
-
-If you wish to have full functionality of the GUIs and are
+Alos note, if you wish to have full functionality of the GUIs and are
```
ntejos: "Also"

profxj:
Merging

PR #359, opened 2017-02-15
LSF for STIS echelle gratings

jnburchett:
Enables LSF retrieval given the wavelength-dependent pixel scale of these modes


linetools/spectra/lsf.py
```diff
@@ -331,7 +334,12 @@ def load_STIS_data(self):
             # reformat rel_pix
             data_aux['rel_pix'] = self.check_and_reformat_relpix(data_aux['rel_pix'])
             # create column with absolute wavelength based on pixel scales
-            wave_aux = float(wa_names[ii])*u.AA + data_aux['rel_pix']*pixel_scale_dict[grating]
+            if not isinstance(pixel_scale_dict[grating],u.quantity.Quantity):
+                # deal with wavelength-dependent pixel scale (i.e., STIS echelle)
+                scalefac = 1./float(pixel_scale_dict[grating].split('/')[-1])
+                wave_aux = float(wa_names[ii])*u.AA * (1. + data_aux['rel_pix']*scalefac)
+            else:
+                wave_aux = float(wa_names[ii])*u.AA + data_aux['rel_pix']*pixel_scale_dict[grating]
```
ntejos: This is not being used, but fine to keep it in.

linetools/spectra/lsf.py
```diff
@@ -450,7 +458,13 @@ def interpolate_to_wv0(self, wv0):
         lsf_vals = Column(name='kernel', data=lsf_vals)
 
         # create column of relative pixel in absolute wavelength
-        wv_array = [(self.pixel_scale * self._data['rel_pix'][i] + wv0*u.AA).value for i in range(len(self._data))]
+        if not isinstance(self.pixel_scale,u.quantity.Quantity):
+            # deal with wavelength-dependent pixel scale (i.e., STIS echelle)
+            scalefac = 1./float(self.pixel_scale.split('/')[-1])
+            wv_array = [(wv0*u.AA * (1. + scalefac * self._data['rel_pix'][i])).value
+                        for i in range(len(self._data))]
```
ntejos: This approach seems fine, but it still assumes that once wv0 is defined, the relative pixels have all the same relative scale. In reality the pixels scale changes all the time.

ntejos: There could be consequences when applying this approach to spectra where delta_wavelength is not constant for VP fitting...

linetools/spectra/tests/test_lsf_use.py
```diff
@@ -16,9 +16,11 @@ def test_interpolate_to_wv0(plot=False, lp='2'):
     wv0 = 1160 * u.AA
     cos_dict = dict(name='COS', grating='G130M', life_position=lp, cen_wave='1309')
     stis_dict = dict(name='STIS', grating='G140L', slit='52x0.2')
+    stis_echelle_dict = dict(name='STIS', grating='E140M', slit='0.2x0.06')
     lsf_cos = LSF(cos_dict)
     lsf_stis = LSF(stis_dict)
-    for lsf in [lsf_cos, lsf_stis]:
+    lsf_stis_echelle = LSF(stis_dict)
```
ntejos: stis_echelle_dict

linetools/spectra/lsf.py
```diff
@@ -113,6 +113,7 @@ def check_and_reformat_relpix(self, relpix):
 
         At the moment this function does the following:
          - Make sure that the number of relative pixels of the LSF is odd integer
+            - Unfortunately, the STIS LSFs provided by STScI have an even number
```
jnburchett: Added a TODO item a few lines down from here.

linetools/spectra/lsf.py
jnburchett: True.  However, at least in my VP fitting code, the LSF is retrieved for each chunk of contiguous pixels included in the fit.  As a result, the change in pixel scale over the length of each chunk should not be large in most cases.  I agree that we'd eventually like to have a more robust fix, but how we could adapt to changing scales over an full spectrum in one LSF call eludes me.

linetools/spectra/tests/test_lsf_use.py
jnburchett: Good catch!

linetools/spectra/lsf.py
```diff
@@ -331,7 +334,12 @@ def load_STIS_data(self):
             # reformat rel_pix
             data_aux['rel_pix'] = self.check_and_reformat_relpix(data_aux['rel_pix'])
             # create column with absolute wavelength based on pixel scales
-            wave_aux = float(wa_names[ii])*u.AA + data_aux['rel_pix']*pixel_scale_dict[grating]
+            if not isinstance(pixel_scale_dict[grating],u.quantity.Quantity):
```
jnburchett: It's actually of type 'unicode', but I've since fixed.

linetools/spectra/lsf.py
```diff
@@ -450,7 +458,13 @@ def interpolate_to_wv0(self, wv0):
         lsf_vals = Column(name='kernel', data=lsf_vals)
 
         # create column of relative pixel in absolute wavelength
-        wv_array = [(self.pixel_scale * self._data['rel_pix'][i] + wv0*u.AA).value for i in range(len(self._data))]
+        if not isinstance(self.pixel_scale,u.quantity.Quantity):
```
jnburchett: fixed

linetools/spectra/lsf.py
jnburchett: Nevermind, added a fix

jnburchett:
I went ahead and added a fix to deal with the asymmetric LSFs provided by STScI for STIS.  Because this is a bit more substantial change than was called for in the review by @ntejos, I would like to request a final look at this piece before merging.  The major changes occur at ines 337-345 of linetools/spectra/lsf.py

Thanks!

linetools/spectra/lsf.py
```diff
@@ -330,8 +333,24 @@ def load_STIS_data(self):
...
+            if data_aux['rel_pix'][len(data_aux['rel_pix']) // 2] == 0.:
+                pass
+            elif data_aux['rel_pix'][(len(data_aux['rel_pix']) // 2) - 1 ] ==0.:
+                data_aux.insert_row(0,data_aux[-1])
+                data_aux['rel_pix'][0]=-data_aux['rel_pix'][-1]
+            elif data_aux['rel_pix'][(len(data_aux['rel_pix']) // 2) + 1 ] ==0.:
+                data_aux.add_row(data_aux[0])
+                data_aux['rel_pix'][-1]=-data_aux['rel_pix'][0]
+
             # create column with absolute wavelength based on pixel scales
```
ntejos: How about we just change the LSFs tables provided by STScI instead?

ntejos: Remove this patch for now and I will fix this minor issue later. Once removed, the coverage/coverall checks should not complain any more.

jnburchett: I suppose this is a philosophical question: I'm more inclined to leave the data tables from STScI intact.  In the spirit of reproducibility, if someone were to download linetools and, with it, these tables from STScI with the exact file names and *nearly* the same data, I see how I would be really confused if trying to understand how the code works and where the data come from.

jnburchett: Clarification: do you mean remove just lines 343-346 or the entire new block of code 337-345?

ntejos: I meant from 337-345. I agree it's a philosophical choice here, but I rather prefer to have a human being (developer) inspecting the LSF tabulated tables that will be ingested in linetools rather than having this piece of code that could easily fail (without raising an error) if the tables are given in the wrong format.

jnburchett:
Wanted to give an update as this PR has been open for a while.  Todd has uncovered a serious issue while fitting lines in the STIS echelle bandpass, and this issue has something to do with the LSF convolution.  I am inclined to leave this PR open until we have tracked down the source of this issue.

ntejos:
I see, thanks for the update. Can you elaborate more on this unexpected behaviour?

jnburchett:
Sure @ntejos , I can elaborate.  The problem arose when Todd was trying to simultaneously fit z~0.9 CII lines at 687 and 1334 \AA (rest frame) and we noticed that the 1334 lines, which are covered by STIS echelle spectra, appeared offset in velocity from the model when fixed to the parameters derived for the 687 lines.  After some surgery, I found that the offset appeared only after convolving with the STIS LSF.  As it turns out, the problem stemmed from the wavelength range fed to LSF.interpolate_to_wv_array() that included some discontinuities in the pixel scale.  I now think the implementation of the echelle LSFs is largely fine...I just need to modify the individual LSF tables per your request, although I've been experimenting with the best way to do that.  Merely adding an extra row to these tables is not a robust solution, I've found.  A more rigorous procedure for improving these tables for our purposes is well underway...

profxj:
@jnburchett , can I help with this?

jnburchett:
Yes, just thinking about this just earlier...If the decision is that we produce our own LSF tables, I want to make sure these are rock solid, that several of us test them beyond a simple github review, and that any potential users are adequately informed about the LSFs they are employing.  There is a small (?) possibility that our tables could be disseminated independently from the repo.  Not to mention, producing LSFs that were compatible with the linetools framework but stayed true to the STScI-provided data was nontrivial, and I still worry about systematics for the unsuspecting user. @ntejos and @profxj , could the 3 of us chat about this once I've reviewed my various development tests?

ntejos:
Sounds like a good plan to me. (I was also wondering about this PR earlier today...)

jnburchett:
Many have come and gone since I opened this guy....

jnburchett:
Update: I spoke with Todd about this, and he has largely allayed my concerns.  I will proceed to make the new LSF tables as suggested by @ntejos .

ntejos:
Sounds good. It would be nice to chat about this anyways before merging.

jnburchett:
Okay, I think the new tables and their treatment in the code is good to go.  The old code is commented out with 'TODO' items marked inline for removal just prior to merging.  I'm happy to chat about this as you wish, @ntejos and/or @profxj

linetools/lists/sets/llist_v1.1.ascii
```diff
@@ -1121,7 +1121,7 @@
 | 1200.7098 |    NI 1200c |    1 |   1 |   0 |   0 |   0 |    0 |   0 |   0 |    0 |
```
ntejos: Changes to this file are unrelated to this PR. Need @profxj to assess whether this is harmful or not.

linetools/spectra/lsf.py
```diff
@@ -112,7 +112,8 @@ def check_and_reformat_relpix(self, relpix):
         """Performs checks and reformating on a relative pixel array.
 
         At the moment this function does the following:
-         - Make sure that the number of relative pixels of the LSF is odd integer
+         - Make sure that the number of relative pixels of the LSF is an odd integer
+            - Unfortunately, the STIS LSFs provided by STScI have an even number
```
ntejos: Remove this docstring comment. You have changed the STIS tables to have a odd number of rows, right?

linetools/spectra/lsf.py
```diff
@@ -127,13 +128,15 @@ def check_and_reformat_relpix(self, relpix):
 
         # odd integer for total number of relative pixels
         n_pix = len(relpix)
-        assert n_pix % 2 != 0, ValueError('LSF tables must be given as odd integers!')
+        if self.instr_config['name'] != 'STIS':  # STIS LSFs have even no. of pixels
```
ntejos: Ditto, remove this.

linetools/spectra/xspectrum1d.py
```diff
@@ -855,6 +855,10 @@ def box_smooth(self, nbox, preserve=True, **kwargs):
                     self.sig[orig_pix], new_npix) / np.sqrt(nbox)
             else:
                 new_sig = None
+            if self.co_is_set:
+                new_co = ltu.scipy_rebin(self.co[orig_pix], new_npix)
+            else:
+                new_co = None
```
ntejos: This bit is unrelated to this PR. Need approval from @profxj.

linetools/lists/sets/llist_v1.1.ascii
jnburchett: Weird, not sure why that happened

linetools/spectra/lsf.py
jnburchett: done

linetools/spectra/lsf.py
jnburchett: done

linetools/spectra/xspectrum1d.py
jnburchett: Don't know or remember why this would have crept in either....but it's important of course!

jnburchett: Not sure why in this PR exactly, but seems to be important fro spectra with defined continuua.

jnburchett: Is this okay @profxj ?

linetools/lists/sets/llist_v1.1.ascii
jnburchett: Is this okay @profxj ?

profxj: It is a bit odd to add these lines to the EUV list, i.e. they
aren't EUV lines.  I'd say it is better to read them in via
the ISM or Strong sets.

linetools/spectra/xspectrum1d.py
profxj: These changes are in the scipy part which is now
broken and deprecated.

PR #369, opened 2017-03-29
Enable list for RA/DEC input

profxj:
As titled.

Solely for the radec_to_coord() method.

No new docs.

### DIFF
--- a/linetools/utils.py
+++ b/linetools/utils.py
@@ -94,7 +94,7 @@ def radec_to_coord(radec):
 
     Parameters
     ----------
-    radec : str or tuple or SkyCoord
+    radec : str or tuple or SkyCoord or list
         Examples:
         'J124511+144523',
         '124511+144523',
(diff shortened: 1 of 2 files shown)


linetools/utils.py
```diff
@@ -143,6 +146,16 @@ def radec_to_coord(radec):
                 raise ValueError("radec must include + or - for DEC")
             newradec = (radec[0:2]+':'+radec[2:4]+':'+radec[4:sign+3] +':'+radec[sign+3:sign+5]+':'+radec[sign+5:])
             coord = SkyCoord(newradec, frame='fk5', unit=(u.hourangle, u.deg))
+    elif isinstance(radec,list):
+        clist = []
```
ntejos: Does the list need to be tuples of coordinates in degrees?

profxj: No

ntejos: I see; I've missed the recursive bit. Looks good then.

profxj:
Merging

PR #371, opened 2017-04-03
Rebin bug

profxj:
Mainly handles rebinning by turning off the kludgy scipy script 
I had been using.  

This is not backwards compatible but I think it is for the best.

### DIFF
--- a/linetools/spectra/tests/test_xspec_utils.py
+++ b/linetools/spectra/tests/test_xspec_utils.py
@@ -111,11 +111,11 @@ def test_addnoise(spec):
 def test_box_smooth(spec):
     # Smooth
     newspec3 = spec.box_smooth(3)
-    np.testing.assert_allclose(newspec3.flux[4000], 0.9650185, rtol=1e-5)
+    np.testing.assert_allclose(newspec3.flux[4000], 0.8429098, rtol=1e-5)
     assert newspec3.flux.unit == u.dimensionless_unscaled
 
     newspec5 = spec.box_smooth(5)
-    np.testing.assert_allclose(newspec5.flux[3000], 1.0405008,rtol=1e-5)
+    np.testing.assert_allclose(newspec5.flux[3000], 0.7609909,rtol=1e-5)
     # Preserve
     newspec5p = spec.box_smooth(5, preserve=True)
 
--- a/linetools/spectra/utils.py
+++ b/linetools/spectra/utils.py
@@ -228,9 +228,9 @@ def rebin(spec, new_wv, do_sig=False, do_co=False, all=False, grow_bad_sig=False
     flux = spec.flux.value
 
     # Deal with nan
-    badf = np.isnan(flux)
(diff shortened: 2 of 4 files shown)


linetools/spectra/utils.py
```diff
@@ -317,6 +317,8 @@ def rebin(spec, new_wv, do_sig=False, do_co=False, all=False, grow_bad_sig=False
             new_sig[bad_new] = 0.
         # Zero out edge pixels -- not to be trusted
         igd = np.where(gd)[0]
+        if len(igd) == 0:  # Should not get here!
+            pdb.set_trace()
```
ntejos: How about raising an error here?

linetools/utils.py
```diff
@@ -156,13 +156,14 @@ def scipy_rebin(a, *args):
...
-    shape = a.shape
+    shape = aa.shape
     lenShape = len(shape)
     factor = np.asarray(shape)/np.asarray(args)
-    evList = ['a.reshape('] + \
+    evList = ['aa.reshape('] + \
              ['args[%d],factor[%d],'%(i,i) for i in range(lenShape)] + \
              [')'] + ['.mean(%d)'%(i+1) for i in range(lenShape)]
     #print ''.join(evList)
+    pdb.set_trace()
```
ntejos: remove this?

linetools/spectra/tests/test_xspec_utils.py
```diff
@@ -111,11 +111,11 @@ def test_addnoise(spec):
...
     newspec3 = spec.box_smooth(3)
-    np.testing.assert_allclose(newspec3.flux[4000], 0.9650185, rtol=1e-5)
+    np.testing.assert_allclose(newspec3.flux[4000], 0.8429098, rtol=1e-5)
     assert newspec3.flux.unit == u.dimensionless_unscaled
 
     newspec5 = spec.box_smooth(5)
-    np.testing.assert_allclose(newspec5.flux[3000], 1.0405008,rtol=1e-5)
+    np.testing.assert_allclose(newspec5.flux[3000], 0.7609909,rtol=1e-5)
     # Preserve
     newspec5p = spec.box_smooth(5, preserve=True)
```
ntejos: Could you also add the missing test for rebin? (i.e. the one that made you realize the bug).

linetools/spectra/utils.py
profxj: yup

linetools/utils.py
profxj: Modified so that this method throws a DeprecationWarning error.

linetools/spectra/tests/test_xspec_utils.py
profxj: I don't think there really was a bug.  Or I can't recall.
Main issue was that the scipy rebin method was no longer working
because of an update in Python.

profxj:
Merging.

PR #376, opened 2017-04-14
Toward deprecating ion_name and name_ion

profxj:
Almost there.  Am sure I'll deprecate it
altogether before long.


profxj:
Merging

PR #382, opened 2017-04-24
AbsComponent: self.reliability added as mandatory attribute

ntejos:
As stated. Test updated accordingly. I believe having this mandatory attribute is useful, particularly to integrate it with IGMGuesses and Joebvp.


linetools/isgm/abscomponent.py
```diff
@@ -54,11 +54,17 @@ class AbsComponent(object):
...
+            'a' - reliable
+            'b' - possible
+            'c' - uncertain
+            'none' - not defined (default)
+    comment : str, optional
         A comment, default is ``
     """
     @classmethod
-    def from_abslines(cls, abslines, stars=None, **kwargs):
+    def from_abslines(cls, abslines, stars=None, reliability='none', **kwargs):
```
profxj: You want a string here or None ?

linetools/isgm/abscomponent.py
```diff
@@ -127,10 +133,20 @@ def from_dict(cls, idict, coord=None, **kwargs):
             radec = SkyCoord(ra=idict['RA']*u.deg, dec=idict['DEC']*u.deg)
         # Init
         #slf = cls(radec, tuple(idict['Zion']), idict['zcomp'], Quantity(idict['vlim'], unit='km/s'),
+
+        # backwards compatibility
+        reliability = 'none'
```
profxj: Could put this in the if statement, if you wanted.

linetools/isgm/abscomponent.py
```diff
@@ -183,7 +205,7 @@ def __init__(self, radec, Zion, zcomp, vlim, Ej=0./u.cm, A=None,
         # Optional
         self.A = A
         self.Ej = Ej
-        self.comment = comment
+        self.comment = str(comment)
```
profxj: This will not be Python2/3 compatible

linetools/isgm/abscomponent.py
ntejos: String, the default will be 'none'.

profxj:
Merging

PR #387, opened 2017-05-18
SelectLineWidget. Init self.line attribute as str.

ntejos:
I found that I needed to explicitly initialize `self.line` in `SelectLineWidget` for not breaking IGMGuesses using such widget (see also PR#108 in pyigm).


profxj:
ok, merging.

PR #395, opened 2017-06-06
Turn off masking

profxj:
Mainly propagates masking choice into collate.

No new docs or tests

### DIFF
--- a/linetools/spectra/utils.py
+++ b/linetools/spectra/utils.py
@@ -127,6 +127,8 @@ def collate(spectra, **kwargs):
     ----------
     spectra : list
       of XSpectrum1D
+    **kwargs : optional
+      Passed to the XSpectrum1D object generated
 
     Returns
     -------
--- a/linetools/spectra/utils.py
+++ b/linetools/spectra/utils.py
@@ -114,7 +114,7 @@ def splice_two(spec1, spec2, wvmx=None, scale=1., chk_units=True):
     return spec3
 
 
-def collate(spectra):
+def collate(spectra, **kwargs):
     """ Generate a single XSpectrum1D instance containing an array of
     spectra from a list of individual XSpectrum1D spectra.
     Each spectrum is padded with extra pixels so that the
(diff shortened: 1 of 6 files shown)


linetools/spectra/utils.py
```diff
@@ -114,7 +114,7 @@ def splice_two(spec1, spec2, wvmx=None, scale=1., chk_units=True):
     return spec3
 
 
-def collate(spectra):
+def collate(spectra, **kwargs):
     """ Generate a single XSpectrum1D instance containing an array of
```
ntejos: Please add in docstring where are the `**kwargs` passed to.

linetools/spectra/xspectrum1d.py
```diff
@@ -766,13 +767,19 @@ def rebin(self, new_wv, all=False, **kwargs):
         all : bool, optional
           Rebin all spectra in the XSpectrum1D object?
           Set masking='none' to have the resultant spectra all be regsitered, but note
-             that there will still be masking
+             that there will still be masking unless otherwise specified in kwargs
```
ntejos: I did not understand the comment regarding setting masking to 'none'. What is the idea here?

Also, check typo `regsitered`.

linetools/spectra/xspectrum1d.py
```diff
@@ -766,13 +767,19 @@ def rebin(self, new_wv, all=False, **kwargs):
...
         -------
         XSpectrum1D of the rebinned spectrum
         """
         from .utils import rebin, collate
+        #
+        if 'masking' in kwargs:
+            masking = kwargs['masking']
+        else:
+            masking = 'edges'
+        #
```
ntejos: Is this necessary? Given the default masking for `XSpectrum1D` (and hence `collate()`), is `masking='edges'` anyways...

linetools/spectra/utils.py
profxj: done

linetools/spectra/xspectrum1d.py
profxj: Now set to none here and in the __init__

This is a somewhat big change, but is the way 
I should have had it and it is likely the way we
will be using the new Spectrum1D object.
 
I hope it doesn't cause trouble to existing scripts,
but there are likely to be some repercussions..

profxj:
merging

PR #396, opened 2017-06-12
ContinuumFitModifications

jnburchett:
toggle initial continuum display in GUI
make optional number of pixels used when guessing knot location by median flux


profxj:
Looks good.  Merging.

PR #397, opened 2017-06-14
Edits to make Xray lines actually accessible

profxj:
Fixes a bug or two.

Added a test

Note that the EUV line list is no longer cached.

### DIFF
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -87,7 +87,7 @@ class LineList(object):
         self.closest = closest
         self.verbose = verbose
 
-        if not use_ISM_table or llst_key not in ('ISM', 'HI', 'Strong', 'EUV'):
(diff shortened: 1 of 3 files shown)


linetools/lists/linelist.py
```diff
@@ -187,9 +187,8 @@ def load_data(self, use_ISM_table=True, tol=1e-3, use_cache=True):
         else:
             # import pdb
             # pdb.set_trace()
-            raise ValueError(
-                'LineList: load_data: Not ready for this LineList name: {:s}'.format(self.list))
-
+            print('LineList: load_data: Not ready for this LineList name: {:s}'.format(self.list))
+            raise ValueError('LineList:  Available options are --  ISM, Strong, EUV, H2, CO, HI, Galaxy')
```
ntejos: remove EUV?

jnburchett: It's not removed, just not cached.  Indeed, to access the x-ray lines, you load EUV linelist.

PR #398, opened 2017-06-20
Remove units in new_sig

profxj:
Small fix to avoid units.

Am worried there is still a units bug 
in the var array but I don't think it will
cause trouble.


profxj:
merging

PR #399, opened 2017-06-28
Arc linelist

ntejos:
- Added forbidden lines for Galaxy LineList that usually show up in gravitational arcs
- Informative error in AbsComponent


profxj:
Looks fine but you may need to update the llist_v1.1.ascii
in sets/

Merge when ready

ntejos:
I'm a bit lost here. Should I ran `add_galaxy_lines()` from scratch or have something more like `add_xray_lines()`, or run both?

PR #401, opened 2017-06-29
lsf: impose n_half to be integer in lsf.py

ntejos:
As stated. This will prevent an error from future.


profxj:
Simple enough.
Merging

PR #403, opened 2017-07-17
Add widget to lt_xspec for selecting between multiple input spectra

profxj:
As titled.

Also adds functionality to dynaimcally set gui size
based on number of pixels in the monitor.

No new docs or tests.

### DIFF
--- a/linetools/guis/spec_widgets.py
+++ b/linetools/guis/spec_widgets.py
@@ -485,7 +485,7 @@ class ExamineSpecWidget(QWidget):
                             dline.analy['spec'] = tspec
                             dline.limits.set(iwv)
                             dline.measure_ew()
-                            mssg = 'Using dummy '+ dline.__repr__()
+                            mssg = 'Using dummy '+ dline.__repr__()+' for the calculation.'
                             mssg = mssg + ' ::  Obs EW = {:g} +/- {:g}'.format(
                                 dline.attrib['EW'].to(mAA), dline.attrib['sig_EW'].to(mAA))
                 # Display values
--- a/linetools/guis/spec_widgets.py
+++ b/linetools/guis/spec_widgets.py
@@ -151,6 +151,7 @@ class ExamineSpecWidget(QWidget):
             spec.airtovac()
         self.orig_spec = spec  # For smoothing
         self.spec = self.orig_spec
+        self.select = spec.select
         self.parent = parent
 
         # determine the filename (if any)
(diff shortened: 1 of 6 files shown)


ntejos:
May I review this?

profxj:
Please do!

linetools/guis/spec_widgets.py
```diff
@@ -467,6 +471,23 @@ def on_key(self, event):
...
+                            #QtCore.pyqtRemoveInputHook()
+                            #pdb.set_trace()
+                            #QtCore.pyqtRestoreInputHook()
+                            dline = AbsLine(wrest, linelist=llist, z=z.value)
+                            tspec = XSpectrum1D.from_tuple((self.spec.wavelength, self.spec.flux, self.spec.sig))
+                            tspec.normalize(lconti)
+                            dline.analy['spec'] = tspec
+                            dline.limits.set(iwv)
+                            dline.measure_ew()
+                            mssg = 'Using dummy '+ dline.__repr__()
```
ntejos: change message to `"Assuming line is "` (?)

ntejos: These two lines seem duplicates from 471 and 472. May be best to take them out of the `if` and `else` statements.

ntejos:
I tested the GUI and everything seems good. I'm not so convinced whether we want to start labeling the spectra from `0` to `n-1`, wouldn't make more sense to start from `1` to `n` instead? 

I also left comments on the code while ago (see above)... not sure if you considered them.

linetools/guis/spec_widgets.py
```diff
@@ -1294,3 +1315,66 @@ def reload(self):
...
+        self.scale = screen_scale
+        self.parent = parent  # Required
+        spec = self.parent.orig_spec
+
+        # Create the list
+        if spec_labels is None:
+            if hasattr(spec, 'labels'):
+                spec_labels = spec.labels
+            else:
+                spec_labels = ['{:d}'.format(ii) for ii in range(spec.nspec)]
```
ntejos: `{:d}'.format(ii)` -> `{:d}'.format(ii+1)` (?)

linetools/guis/spec_widgets.py
profxj: This is so a user can measure the EW without setting a line.
Have made a small edit.

linetools/guis/spec_widgets.py
profxj: I think I prefer Python indexing.  That is the way
the spectra are truly indexed anyhow.

profxj:
Merging

PR #406, opened 2017-08-28
Turning off z warning in LineLimit

profxj:
And turns one on when instantiating an AbsLine
without an input redshift.

No new docs or tests


profxj:
Merging

PR #408, opened 2017-09-05
Catch all for GUI key stroke errors

profxj:
A bit aggressive, but a nice way to stop PyQt crashing
on a bad key-stroke entry.

I don't think there is a downside.  We could include key
clicks and also add to pyigm (e.g. igmguesses).

### DIFF
--- a/linetools/analysis/interactive_plot.py
+++ b/linetools/analysis/interactive_plot.py
@@ -86,10 +86,14 @@ S,U          Smooth/unsmooth spectrum
         """
         # Requiring inaxes for all of these now
         if (event.key in self.nav_dict['nav']) and event.inaxes:
-            ltgu.navigate(self.nav_dict, event, flux=self.fl, wave=self.wa)
-            self.ax.set_xlim(self.nav_dict['x_minmax'])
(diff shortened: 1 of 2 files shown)


profxj:
@ntejos -- ok with this?

linetools/analysis/interactive_plot.py
```diff
@@ -86,10 +86,14 @@ def on_keypress_navigate(self, event):
...
         if (event.key in self.nav_dict['nav']) and event.inaxes:
-            ltgu.navigate(self.nav_dict, event, flux=self.fl, wave=self.wa)
-            self.ax.set_xlim(self.nav_dict['x_minmax'])
-            self.ax.set_ylim(self.nav_dict['y_minmax'])
-            self.fig.canvas.draw()
+            try:
+                ltgu.navigate(self.nav_dict, event, flux=self.fl, wave=self.wa)
+            except KeyError:
+                pass
+            else:
```
ntejos: What is this `else` doing?

profxj: just continuing the code if we don't hit the error.

profxj:
merging.

PR #410, opened 2017-09-18
Unique components

profxj:
There are two threads to this PR:

- Refactor from LineLimits to zLimits
- Add a unique_components() method

The first allows for limits without reference to wrest.
That becomes optional, so previous functionality is maintained.

The second is a method useful for querying lists of components.

I also removed GaussianAbsorption1D as that is deprecated in astropy.

Includes a new test, but no new docs.

### DIFF
--- a/linetools/analysis/zlimits.py
+++ b/linetools/analysis/zlimits.py
@@ -16,6 +16,8 @@ ckms = const.c.to('km/s')
 
 class zLimits(object):
     """ An object for handling the 'limits' of a line
+    The input redshift is not meant to be modified (ever).
+    One should re-instantiate instead.
 
     Properties
     ----------
--- a/linetools/analysis/zlimits.py
+++ b/linetools/analysis/zlimits.py
@@ -1,4 +1,4 @@
-"""Module containing the LineLimits object
+"""Module containing the zLimits object
 """
 from __future__ import print_function, absolute_import, division, unicode_literals
 
@@ -13,7 +13,8 @@ from linetools import utils as ltu
 
 ckms = const.c.to('km/s')
 
-class LineLimits(object):
+
+class zLimits(object):
     """ An object for handling the 'limits' of a line
 
     Properties
@@ -23,7 +24,7 @@ class LineLimits(object):
     zlim : tuple of floats
       Redshift limits for a line
       Defined as wave/wrest - 1.
-    wvlim : Quantity array
+    wvlim : Quantity array, only if _wrest is set
       wavelength limits for the line in observer frame
     vlim : Quantity array
       velocity limits for the line
@@ -41,7 +42,7 @@ class LineLimits(object):
         if not isinstance(aline, (AbsLine, EmLine)):
             raise IOError("Input aline must be AbsLine or EmLine")
         #
-        slf = cls(aline.wrest, z, zlim)
+        slf = cls(z, zlim, wrest=aline.wrest)
         return slf
 
     @classmethod
@@ -57,31 +58,35 @@ class LineLimits(object):
         -------
 
         """
-        wrest = ltu.convert_quantity_in_dict(idict['wrest'])
-        slf = cls(wrest, idict['z'], idict['zlim'])
+        try:
+            wrest = ltu.convert_quantity_in_dict(idict['wrest'])
+        except KeyError:
+            wrest = None
+        slf = cls(idict['z'], idict['zlim'], wrest=wrest)
         # Return
         return slf
 
-    def __init__(self, wrest, z, zlim, **kwargs):
+    def __init__(self, z, zlim, wrest=None, **kwargs):
         """
         Parameters
         ----------
-        wrest : Quantity
-          Rest wavelength of the line.
         z : float
           Redshift of the line.
         zlim : tuple or list
           Redshift limits for a line
           Defined as wave/wrest - 1.
           Ok to have zlim[1]==zlim[0], but then self.is_set() == False
+        wrest : Quantity, optional
+          Rest wavelength.  Used for SpectralLine objects
         """
         # Error checking
         if not isinstance(z, (float,int)):
             raise IOError("Input z must be a float")
         if not isinstance(zlim, (tuple,list)):
             raise IOError("Input zlim must be a tuple or list")
-        if not isinstance(wrest, Quantity):
-            raise IOError("Input wrest must be a quantity")
+        if wrest is not None:
+            if not isinstance(wrest, Quantity):
+                raise IOError("Input wrest must be a quantity")
 
         # Data
         self._data = {}
@@ -124,7 +129,8 @@ class LineLimits(object):
         """ Update all the values
         """
         #self._data['zlim'] = self._zlim
-        self._wvlim = self._wrest*(1+np.array(self._zlim))
+        if self._wrest is not None:
+            self._wvlim = self._wrest*(1+np.array(self._zlim))
         self._vlim = ltu.dv_from_z(self._zlim, self._z)
 
     def is_set(self):
@@ -198,7 +204,8 @@ class LineLimits(object):
         ldict['zlim'] = self.zlim
         for key in ['wvlim', 'vlim', 'wrest']:
             obj = getattr(self, key)
-            ldict[key] = dict(value=obj.value,
+            if obj is not None:
+                ldict[key] = dict(value=obj.value,
                               unit=obj.unit.to_string())
         # Return
         return ldict
@@ -206,10 +213,11 @@ class LineLimits(object):
     def __repr__(self):
         txt = '<{:s}'.format(self.__class__.__name__)
         # wrest
-        txt = txt + ' wrest={:g}'.format(self.wrest)
         txt = txt + ' z={:g}'.format(self.z)
         txt = txt + ' zlim={}'.format(self.zlim)
-        txt = txt + ' wvlim={}'.format(self.wvlim)
+        if self._wrest is not None:
+            txt = txt + ' wrest={:g}'.format(self.wrest)
+            txt = txt + ' wvlim={}'.format(self.wvlim)
         txt = txt + ' vlim={}'.format(self.vlim)
         txt = txt + '>'
         return (txt)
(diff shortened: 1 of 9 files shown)


linetools/isgm/utils.py
```diff
@@ -506,23 +506,25 @@ def synthesize_components(components, zcomp=None, vbuff=0*u.km/u.s):
...
     # zcomp
     if zcomp is None:
         zcomp = np.mean([comp.zcomp for comp in components])
-    synth_comp.zcomp = zcomp
+
     # Set vlim by min/max  [Using non-relativistic + buffer]
     vmin = u.Quantity([(comp.zcomp-zcomp)/(1+zcomp)*const.c.to('km/s')+comp.vlim[0] for comp in components])
     vmax = u.Quantity([(comp.zcomp-zcomp)/(1+zcomp)*const.c.to('km/s')+comp.vlim[1] for comp in components])
-    synth_comp.vlim = u.Quantity([np.min(vmin)-vbuff, np.max(vmax)+vbuff])
+    vlim = u.Quantity([np.min(vmin)-vbuff, np.max(vmax)+vbuff])
```
jnburchett: Looks like the default value of vbuff is 0 km/s but the docstring indicates 10 km/s?

linetools/isgm/utils.py
```diff
@@ -886,3 +888,58 @@ def joebvp_from_components(comp_list, specfile, outfile):
     f.close()
 
 
+def unique_components(comps1, comps2, tol=5*u.arcsec):
+    """ Identify which members of comps1 are *not* within
+    comps2, to given tolerances.  Note, that repeats in
+    comps1 are not looked for.
```
jnburchett: Perhaps clarify 2nd sentence? "Note: nonunique members of comps1 are not identified."

linetools/analysis/zlimits.py
```diff
@@ -23,7 +24,8 @@ class LineLimits(object):
     zlim : tuple of floats
       Redshift limits for a line
       Defined as wave/wrest - 1.
-    wvlim : Quantity array
+      or
+    wvlim : Quantity array, only if _wrest is set
       wavelength limits for the line in observer frame
```
ntejos: this docstring `only if _wrest is set` is confusing. Maybe say that this is an optional parameter that applies for zLimits to individual AbsLines (for which a single wrest makes sense)?

linetools/analysis/zlimits.py
```diff
@@ -23,7 +24,8 @@ class LineLimits(object):
     zlim : tuple of floats
```
ntejos: add `z` to docstring

linetools/analysis/zlimits.py
```diff
@@ -57,31 +59,35 @@ def from_dict(cls, idict):
...
 
-    def __init__(self, wrest, z, zlim, **kwargs):
+    def __init__(self, z, zlim, wrest=None, **kwargs):
         """
         Parameters
         ----------
-        wrest : Quantity
-          Rest wavelength of the line.
         z : float
           Redshift of the line.
```
ntejos: redshift of the line -> redshift to apply the zlimits to

linetools/analysis/zlimits.py
```diff
@@ -57,31 +59,35 @@ def from_dict(cls, idict):
...
         zlim : tuple or list
           Redshift limits for a line
           Defined as wave/wrest - 1.
           Ok to have zlim[1]==zlim[0], but then self.is_set() == False
+        wrest : Quantity, optional
+          Rest wavelength.  Used for SpectralLine objects
```
ntejos: Used for SpectralLine objects -> Used for applying zLimits to individual SpectralLine objects

linetools/analysis/zlimits.py
```diff
@@ -142,7 +149,7 @@ def is_set(self):
             return False
 
     def set(self, inp, chk_z=False):#, itype='zlim'):
-        """ Over-ride = to re-init values
+        """ Set (or reset) limits relative to self._z
```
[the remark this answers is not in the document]
ntejos: agreed with the suggestion above.

linetools/isgm/abscomponent.py
```diff
@@ -207,8 +208,9 @@ def __init__(self, radec, Zion, zcomp, vlim, Ej=0./u.cm, A=None,
         # Required
         self.coord = ltu.radec_to_coord(radec)
         self.Zion = Zion
-        self.zcomp = zcomp
-        self.vlim = vlim
+        # Limits
+        zlim = ltu.z_from_dv(vlim, zcomp)
+        self.limits = zLimits(zcomp, zlim.tolist())
```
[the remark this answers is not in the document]
ntejos: Good idea, should probably be a method of `zLimit` class.

linetools/isgm/utils.py
ntejos: I do not understand this either. How can a component be _within_ another component? This is based on velocity limits or was supposed to identify two components that may be considered the same physical entity?

linetools/analysis/zlimits.py
```diff
@@ -23,7 +24,8 @@ class LineLimits(object):
     zlim : tuple of floats
       Redshift limits for a line
       Defined as wave/wrest - 1.
-    wvlim : Quantity array
+      or
```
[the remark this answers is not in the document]
profxj: I even removed the line above as wrest is not required

linetools/analysis/zlimits.py
```diff
@@ -41,7 +43,7 @@ def from_specline(cls, aline, z, zlim):
         if not isinstance(aline, (AbsLine, EmLine)):
             raise IOError("Input aline must be AbsLine or EmLine")
         #
-        slf = cls(aline.wrest, z, zlim)
+        slf = cls(z, zlim, wrest=aline.wrest)
```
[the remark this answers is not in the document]
profxj: added

linetools/analysis/zlimits.py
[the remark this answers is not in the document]
profxj: edited

linetools/analysis/zlimits.py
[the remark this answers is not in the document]
profxj: edited

linetools/isgm/abscomponent.py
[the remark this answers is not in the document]
profxj: I understand the desire, but how does one modify the limits
in that case?  Leave them as is?  

I think I'd argue that the user should in that case instantiate
a new zLimits object.

linetools/isgm/utils.py
profxj: Must have used to have been 10 km/s.  I guess
I figured that was a problem..

Have edited the doc string accordingly and kept it at 0 km/s.

linetools/isgm/utils.py
profxj: have tried to clarify.

linetools/analysis/zlimits.py
profxj: done

linetools/analysis/zlimits.py
profxj: improved

linetools/analysis/zlimits.py
profxj: edited.  ps.  your version looks a bit out of date

linetools/analysis/zlimits.py
profxj: edited

profxj:
Ok, the outstanding issue is how to modify the redshift if at all.

I am preferring that the user re-instantiate, i.e. self._z is *fixed*.
But you can convince me otherwise.

Another option would be a method that takes in a new redshift and
shifts the limits accordingly.  It would also reinstantiate, e.g.

self.limits = self.limits.shift_z(z)

Thoughts?

ntejos:
Now that I re-think about it, `zLimits` is a utility class to be used within the framework of `SpectralLines`, `AbsComponents`, etc. It is not meant for stand-alone use (let me know if you envision where this would be useful). In that context, it may be preferable (cleaner) that for changing the redshift, one has to create a new object from scratch. The advantage for the alternative is speed, but it is not obvious what to do with the limits when `self._z` changes. Thus, cleaner may be better than faster in this context.

linetools/isgm/utils.py
ntejos: This remains unclear to me. I beleive you try to identify component duplications (?), but in this case why to allow for a tolerance? If you allow a tolerance (specially in velocity limits), then what would happen for multiple contiguous and overlapping components? Will there be considered the same? If so, why is this useful?

profxj: Oops, I forgot to push utils.py

Look at the new docstring please

jnburchett:
Yeah, I'm not married to the idea of modifying the z of an existing zLimits object if you guys feel strongly that the user should create a new one. However, at the very least can we inform the user that they need to reinstantiate for a different redshift?

As for what to do with the limits told we did change it, one basic option is to keep the vlim and adjust zlim accordingly. I see that I'm able to change the z of an AbsLine with .setz(), so the point is moot in most applications I suppose.  Here the user is advised to update limits but not required to. If I change the vlim after using setz(), the zlim are updated as well, so I suppose what I foresee doing is possible given current functionality.

-Joe

linetools/isgm/utils.py
```diff
@@ -889,9 +889,10 @@ def joebvp_from_components(comp_list, specfile, outfile):
 
 
 def unique_components(comps1, comps2, tol=5*u.arcsec):
-    """ Identify which members of comps1 are *not* within
-    comps2, to given tolerances.  Note, that repeats in
-    comps1 are not looked for.
+    """ Identify which AbsComponent members of the comps1 list
+    are *not* within the comps2 list, to given tolerances.
+    Note, AbsComponent objects in the comps1 list are examined
```
jnburchett: AbsComponent objects in the comps1 list are *not* examined against each other right?  I though that's what the previous docstring was saying.

profxj: whoops..  edited

profxj:
Ok, give this a thumbs up (if you agree).

ntejos:
I still do not understand the idea behind unique_components(), but I guess is fine.

profxj:
Merging.

@ntejos -- You'll see this new method in use shortly.  (z1IGM)

Note, I suspect igmguesses will break again.  But I'll help fix it.

PR #411, opened 2017-09-21
Cleaner look for lt_radec

profxj:
As advertised.

Small changes for a somewhat cleaner look.

### DIFF
--- a/linetools/scripts/lt_radec.py
+++ b/linetools/scripts/lt_radec.py
@@ -18,7 +18,7 @@ def parser(options=None):
     import argparse
     # Parse
     parser = argparse.ArgumentParser(
-        description='Print coordinates in several formats from input one.')
+        description='Print coordinates in several formats from input one. [v1.1]')
     parser.add_argument("inp", nargs='?', default=None, help="RA,DEC (e.g. 152.25900,7.22885), JXX (e.g. J100902.16+071343.8)")
     parser.add_argument("--epoch", default=2000., type=float, help="Epoch [Not functional]")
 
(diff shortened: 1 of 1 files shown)


linetools/scripts/lt_radec.py
```diff
@@ -43,10 +43,15 @@ def main(args=None):
     coord = ltu.radec_to_coord(icoord)
 
     # Time to print
+    print('===================================')
     print('J{:s}{:s}'.format(coord.ra.to_string(unit=u.hour,sep='',pad=True),
```
ntejos: I would remove this "===== ..." header, as no every terminal has monospace fonts.

linetools/scripts/lt_radec.py
```diff
@@ -43,10 +43,15 @@ def main(args=None):
...
                              coord.dec.to_string(sep='',pad=True,alwayssign=True)))
+    print('   ')
+    print('   {:s} {:s}'.format(coord.ra.to_string(unit=u.hour,sep=':',pad=True),
+                             coord.dec.to_string(sep=':',pad=True,alwayssign=True)))
```
ntejos: Say explicitly that these are J2000 coordinates somewhere?

linetools/scripts/lt_radec.py
profxj: changed to a blank line (helps to have that separation)

linetools/scripts/lt_radec.py
profxj: added.  good suggestion.

profxj:
merging

PR #412, opened 2017-09-25
A few AbsComponent edits

profxj:
Edits mainly to speed up reading JSON files.

Also turned off 'stars' as a required parameter when
generating a fine-structure component.  Defaults
to a single star if not provided.

Failing one Travis test which is not related to linetools.


linetools/isgm/utils.py
```diff
@@ -162,7 +162,8 @@ def build_components_from_dict(idict, coord=None, **kwargs):
             if isinstance(idict['lines'][key], AbsLine):
                 line = idict['lines'][key]
             elif isinstance(idict['lines'][key], dict):
-                line = AbsLine.from_dict(idict['lines'][key], coord=coord)
+                pdb.set_trace()
```
jnburchett: Why is this here?

profxj:
merging

PR #415, opened 2017-10-03
AbsSightline from_systems()

jnburchett:
Includes a test


profxj:
Looks good to me.

Merging.

PR #418, opened 2017-10-09
Read joeb VP output file

profxj:
As titled.

Returns a list of AbsComponents but @jnburchett or @ntejos 
may want more.

Test but no docs.


linetools/isgm/io.py
```diff
@@ -41,3 +50,67 @@ def abssys_from_json(filename):
...
+    # init
+    if llist is None:
+        llist = LineList('ISM')
+    comps = []
+    # Read
+    vp_data = Table.read(filename, format='ascii')
+    nflags, nidx = np.unique(vp_data['nflag'], return_index=True)
+
+    # Subset by nflag; Build components
+    for nflag in nflags:
```
jnburchett: I think we shouldn't pick out components by the 'nflag' values.  I see the logic and while this will work in many cases, 'nflag' can also take on values 0 (let parameter vary without being tied to other lines) or 1 (fix parameter value).  My approach chooses components based on 'zsys' and then 'trans' (which should probably be called 'species', but anyway...).  However, this isn't perfect either, because one may already have declared multiple components with the same systemic redshift but different velocities during the fit.  The best approach might involve employing the clustering analysis to group different transitions of the same species and redshift by velocity.

linetools/isgm/io.py
```diff
@@ -41,3 +50,67 @@ def abssys_from_json(filename):
...
+        mt_lines = np.where(vp_data['nflag'] == nflag)[0]
+        if len(np.unique(vp_data['vel'][mt_lines])) != 1:
+            pdb.set_trace()
```
jnburchett: This is immediately tripped by the CASBaH parameter file I tried loading because we will occasionally let the velocity centroid of a line flop around a little in the fit due to the COS wavelength uncertainty.  Another reason the clustering analysis is perhaps useful here.

linetools/isgm/io.py
```diff
@@ -41,3 +50,67 @@ def abssys_from_json(filename):
...
+        z_fit = vp_data['zsys'][mt_lines[0]] + vp_data['vel'][mt_lines[0]] * (1 + vp_data['zsys'][mt_lines[0]]) / ckms
+        # Loop on abs lines
+        alines = []
+        for idx in mt_lines:
+            zlim = [vp_data['zsys'][idx] +
+                    vp_data[vkey][idx] * (1 + vp_data['zsys'][idx]) / ckms
+                    for vkey in ['vlim1', 'vlim2']]
+            absline = AbsLine(vp_data['restwave'][idx] * u.AA, z=z_fit, zlim=zlim, linelist=llist)
```
jnburchett: You could just declare the limits in velocity space too, right?

linetools/isgm/io.py
```diff
@@ -41,3 +50,67 @@ def abssys_from_json(filename):
...
+            # Add measurements [JB -- Want to capture anything else??]
+            absline.attrib['coord'] = coord
+            absline.attrib['flag_N'] = 1
+            absline.attrib['logN'] = vp_data['col'][idx]
+            absline.attrib['sig_logN'] = vp_data['sigcol'][idx]
+            absline.attrib['b'] = vp_data['bval'][idx]
+            absline.attrib['sig_b'] = vp_data['sigbval'][idx]
+            absline.attrib['z'] = z_fit
+            absline.attrib['sig_z'] = (1 + vp_data['z_comp'][idx]) * vp_data['sigvel'][idx] / ckms
+            absline.attrib['specfile'] = vp_data['specfile'][idx]
```
jnburchett: What's in the 'specfile' is probably not going to be especially useful, since it records the path to file at the time of the fit.  What *is* useful (and even necessary for some of my analysis) is setting absline.analy['spec'].   For this reason, the function should allow the user to specify the spectrum, either an XSpectrum1D object or a path to the file.

linetools/isgm/io.py
```diff
@@ -41,3 +50,67 @@ def abssys_from_json(filename):
...
+        for idx in mt_lines:
+            zlim = [vp_data['zsys'][idx] +
+                    vp_data[vkey][idx] * (1 + vp_data['zsys'][idx]) / ckms
+                    for vkey in ['vlim1', 'vlim2']]
+            absline = AbsLine(vp_data['restwave'][idx] * u.AA, z=z_fit, zlim=zlim, linelist=llist)
+            # Add measurements [JB -- Want to capture anything else??]
+            absline.attrib['coord'] = coord
+            absline.attrib['flag_N'] = 1
+            absline.attrib['logN'] = vp_data['col'][idx]
+            absline.attrib['sig_logN'] = vp_data['sigcol'][idx]
```
jnburchett: absline.attrib['N'] and 'sig_N' should also be set, as these are needed for AbsComponent.synthesize_column()

linetools/isgm/io.py
```diff
@@ -41,3 +50,67 @@ def abssys_from_json(filename):
 
     # Return
     return abs_sys
+
+
+def read_joebvp(filename, coord, llist=None):
```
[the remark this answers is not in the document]
profxj: added.  optional for now.

linetools/isgm/io.py
profxj: egads, I worry about the clustering.

Let me try zsys + trans.

linetools/isgm/io.py
profxj: gack.  am still resisting to the clustering.
as a human, we had every intent of grouping
these lines together as a component, right?

profxj: anyways, have made this an optional check and
defaulted to False

linetools/isgm/io.py
profxj: I don't think so.  Not for a SpectralLine
But we could add that sometime.

linetools/isgm/io.py
profxj: handled

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
...
+        abscomp.attrib = alines[0].attrib.copy()
+        # Remove undesired keys
+        for key in ['EW', 'sig_EW', 'flag_EW', 'N', 'sig_N']:
+            abscomp.attrib.pop(key)
+        # And more
+        for key in ['flag_N', 'logN', 'sig_logN']:
+            setattr(abscomp, key, abscomp.attrib[key])
+        # Errors must be in first line!
+        assert abscomp.sig_logN > 0.
+        comps.append(abscomp)
```
ntejos: "reliability" and "comment" columns should also be included.

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
 
     # Return
     return abs_sys
+
+
+def read_joebvp(filename, coord, llist=None, specfile=None, chk_vel=False):
+    """
```
ntejos: Wasn't a version of this already coded in package `joebvp`?

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
...
+    Parameters
+    ----------
+    filename : str
+      joeB VP filename
+    coord : SkyCoord
```
ntejos: Explain coord is supposed to be the coordinate of the QSO sightline for the `AbsComponents`...

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
...
+    llist : LineList, optional
```
ntejos: Explain how `llist` will be used if given, and what happens otherwise.

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
...
+    specfile : str, optional
```
ntejos: Explain specfile is the name of the joebvp file.

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
...
+    lbls = np.array(lbls)
+    ulbls = np.unique(lbls)
+
+    # Subset by nflag; Build components
+    for lbl in ulbls:
+        mt_lines = np.where(lbls == lbl)[0]
+        if chk_vel:
+            if len(np.unique(vp_data['vel'][mt_lines])) != 1:
+                pdb.set_trace()
+        z_fit = vp_data['zsys'][mt_lines[0]] + vp_data['vel'][mt_lines[0]] * (1 + vp_data['zsys'][mt_lines[0]]) / ckms
```
ntejos: use `ltu.z_from_v()` instead.

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
...
+            absline.attrib['coord'] = coord
+            absline.attrib['flag_N'] = 1
+            absline.attrib['logN'] = vp_data['col'][idx]
+            absline.attrib['sig_logN'] = vp_data['sigcol'][idx]
+            absline.attrib['b'] = vp_data['bval'][idx]
+            absline.attrib['sig_b'] = vp_data['sigbval'][idx]
+            absline.attrib['z'] = z_fit
+            absline.attrib['sig_z'] = (1 + vp_data['z_comp'][idx]) * vp_data['sigvel'][idx] / ckms
+            if specfile is None:
+                absline.attrib['specfile'] = vp_data['specfile'][idx]
```
ntejos: Is this something we want? If the joebvp file already has the specfile, why should we allow the user to change it?

linetools/isgm/tests/files/group_9.VP
```diff
@@ -0,0 +1,4 @@
+specfile|restwave|zsys|col|sigcol|bval|sigbval|vel|sigvel|nflag|bflag|vflag|vlim1|vlim2|wobs1|wobs2|pix1|pix2|z_comp|trans
```
ntejos: Optional columns "comment" and "reliability" may exist.

linetools/isgm/io.py
```diff
@@ -41,3 +51,83 @@ def abssys_from_json(filename):
 
     # Return
     return abs_sys
+
+
+def read_joebvp(filename, coord, llist=None, specfile=None, chk_vel=False):
```
ntejos: Name should be something like: `components_from_joebvp()` (Note that we already have `isgm.utils.joebvp_from_components()`. Maybe worth using both in the test case.

profxj: renamed a few things as suggested

linetools/isgm/io.py
profxj: I don't know.  @jnburchett ?

linetools/isgm/io.py
profxj: added

linetools/isgm/io.py
profxj: added

linetools/isgm/io.py
profxj: good call

linetools/isgm/io.py
profxj: @jnburchett had a good reason (see above)

linetools/isgm/io.py
profxj: I'm not sure how to do this.  Going to leave for @ntejos

linetools/isgm/tests/files/group_9.VP
profxj: same

ntejos:
Have added a commit for reliability and comments. Please review. Also, it seems that `"z_fit"` is the same as `"z_comp"`. If this is the case, would be good to leave only one of them in the code.

profxj:
Merging

PR #421, opened 2017-10-16
Voigt model

profxj:
Fixes for new astropy.

### DIFF
--- a/linetools/analysis/tests/test_voigt.py
+++ b/linetools/analysis/tests/test_voigt.py
@@ -3,11 +3,46 @@ from __future__ import print_function, absolute_import, division, unicode_litera
 import numpy as np
 import pytest
 import pdb
+
 from astropy import units as u
+from astropy import constants as const
 
 from linetools.spectralline import AbsLine
 from linetools.analysis import voigt as lav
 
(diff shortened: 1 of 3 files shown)


linetools/analysis/tests/test_voigt.py
```diff
@@ -3,11 +3,46 @@
...
+    fitvoigt.logN.min = 12.
+    fitvoigt.b.min = 10.
+    fitvoigt.z.min = 2. + -100. * (1 + 2.) / c_kms
+    fitvoigt.z.max = 2. + 100 * (1 + 2.) / c_kms
+
+    # Fit
+    fitter = fitting.LevMarLSQFitter()
+    parm = fitter(fitvoigt,vmodel.wavelength[abslin.analy['pix']].value,
+                  vmodel.flux[abslin.analy['pix']].value)
+    assert np.abs(parm.logN.value-np.log10(abslin.attrib['N'].value)) < 0.1
```
jessicawerk: So, if this fails, then what happens? We don't use it?

profxj: ask @ntejos over a pisco sour  ;)

profxj:
merging.

PR #422, opened 2017-10-18
Transition lookup by Zion tuple

jnburchett:
Make LineList.all_transitions and LineList.strongest_transitions accept Zion tuple, e.g., (8,6)

### DIFF
--- a/linetools/lists/linelist.py
+++ b/linetools/lists/linelist.py
@@ -571,12 +571,14 @@ class LineList(object):
 
         Parameters
         ----------
-        line : str or Quantity
+        line : str or Quantity or tuple
             Name of line. (e.g. 'HI 1215', 'HI', 'CIII', 'SiII',
             1215.6700*u.AA). When string contains spaces it only
             considers the first part of it, so 'HI' and 'HI 1215' and
             'HI 1025' are all equivalent. To retrieve an unknown line
             use string 'unknown'.
+
+            If tuple -- Zion, e.g. (6,2)
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


linetools/lists/linelist.py
```diff
@@ -595,6 +597,9 @@ def all_transitions(self, line, debug=False):
         elif isinstance(line, Quantity):  # Rest wavelength (with units)
             data = self.__getitem__(line)
             return self.all_transitions(data['name'])
+        elif isinstance(line, tuple): # Zion input
+            data = self.__getitem__(line)
+            return self.all_transitions(data['name'][0].split(' ')[0])
```
ntejos: This seems duplication of code. You could simply replace current line 597: 

`elif isinstance(line, Quantity):` -> `elif isinstance(line, (tuple, Quantity)):`.

jnburchett: Fair enough, but then I still have test whether 'line' is a tuple because of the differences between lines 599 and 602.

ntejos: The spliting of name should be done in current line 596...

jnburchett: Only if all_transitions returns just 1 transition...

ntejos: Oh, I see. I would prefer avoid duplication of code, so maybe the solution is to have the code check whether the output is `dict` (implying 1 transition) or `Table` (implying >1 transition)

jnburchett:
We cool, @profxj and @ntejos ?

profxj:
Yup

PR #423, opened 2017-10-20
Update lsf.py

rcooke-ast:
Needs to be integer division


ntejos:
Looks good to me, will merge after travis is done checking tests.

PR #431, opened 2017-11-18
Line list updates

profxj:
Fix SiIV and CIV 

Also add Ref to lt_line output


profxj:
Address Issue #431

profxj:
I've checked the code.  The hard-coded tolerance is 0.001

So the SiIV example would be a **miss** for Morton 2003 
and a hit for Verner 94.  Verner 94 returns the atomic data,
it is just that its f-value is not **identical** to Morton's.

@yzhenggit -- please review this code and merge if you are happy.
Thanks.

PR #433, opened 2017-12-04
cashman17.ascii

fcashman:
I've created cashman17.ascii, it is able to merge. Take a look and see if it's okay.


profxj:
Looks good.

Merging

PR #434, opened 2017-12-04
Update parse.py to include "read_cashman17()"

fcashman:
File placed after "def read_galabs():", I wasn't sure where to put it. Let me know if this is okay and if I need to modify anything.

### DIFF
--- a/linetools/lists/parse.py
+++ b/linetools/lists/parse.py
@@ -367,6 +367,31 @@ def read_galabs():
     return data
 
 
+def read_cashman17():
+    """ read lines from Cashman et al. 2017
+    
+    Returns 
+    -----------
+    Modified table from Cashman et al. 2017
+    """
+    cashman17_fil = lt_path + '/data/lines/cashman17.ascii'
+    print('linetools.lists.parse: Reading linelist --- \n   {:s}'.format(cashman17_fil))
+    data = Table.read(cashman17_fil, format='ascii', guess=False, comment=';',delimiter=',')
+
+    # Remove
+    data.remove_column('lower_LSJ_level')
+    data.remove_column('lower_LSJ_level')
+    data.remove_column('wobs')
+    data.remove_column('log(g*f)')
+    data.remove_column('grade')
+
+    # Units
+    data['wrest'].unit = u.AA
+
+    # Return
+    return data
+
+
 def mask_gal(data, mask_keys=None):
     """Masks linelist attributes for all galaxy lines
 


profxj:
Looks good.  But we also need the Table: cashman17.ascii

And are you recommending this supersede anything that
came before it (which is fine)?

Last question:  Do your rest wavelengths match those of 
previous works e.g. Morton 2003?  Or are those different?
I fear part of my implementation is to use the lines in the
sets files, i.e.

https://github.com/linetools/linetools/blob/master/linetools/lists/sets/llist_v1.2.ascii

profxj:
Ah, I see now the ASCII file was in a separate PR (now merged).

But please answer the other queries.

Thanks!

fcashman:
We are recommending that these data supersede anything that came before it. The rest wavelengths, like Morton's, are Ritz wavelengths derived from the published energy values. However, as it had been 14 years, we chose to update the wavelengths to the most recently adopted NIST values, though the wavelength changes are not drastic.

profxj:
Got it @fcashman 

I don't suppose I can ask you to do this bit of book-keeping:

cross-match your wavelengths against the llist_v1.2 file I linked above
and provide me a list of all those that don't match to a very small tolerance,
e.g. 1e-3Ang?

I can do this myself of course but it will be faster if you do so.

fcashman:
Sure, I'd be glad to! :-)

profxj:
Thanks!

ntejos:
Hi,

Just a comment regarding LineList that may (or may not) be relevant to this PR. For some reason, some transitions that are already in Morton or Verner catalogs are not passed down to the LineList object. For example, I was looking at S IV 1073 which is in both (Morton and Verner) but not in ISM LineList. This probably has to do with the parsing of each file.

profxj:
Am pretty sure this is because the line is not in the
llists_v1.2 file.

Feel free to add it and/or start a v1.3

ntejos:
Indeed, it was not a parsing problem but just that it was not in the llist set v1.2. Will start a v1.3 soon.

fcashman:
I finished cross-checking the wavelengths against the llist_v1.2 file, and attached the list here for you to take a look at. I kept your same naming convention, which I wish I had noticed before I made my .ascii file. During this process, I created another file of lines/wavelengths from my paper that are not in list_v1.2. Would this be helpful? These match the naming convention in cashman17.ascii, but can be modified.
[wave_update_llist_v1.2.txt](https://github.com/linetools/linetools/files/1556563/wave_update_llist_v1.2.txt)

profxj:
Merging.

ntejos:
@fcashman Regarding wave_update_llist_v1.2.txt, I see several lines that are already in v1.2, at least the first 5, and I suspect this trend will continue. Are you sure this file is the correct one? Thanks!

fcashman:
@ntejos Yes, this is the list of lines from v1.2 who wavelengths need updating (lines with changes in 1e-3Ang place). Every one of these lines should be in v1.2. I generated another list of lines that aren't in v1.2, but have questions about naming. I want to follow the naming convention I see in v1.2, but then these lines won't match how I named them in cashman17.ascii. Does this matter?

ntejos:
I see now. This is a bit more complex that what I thought. For now, I will only update the wavelengths in the llist_v1.3 to match those provided by you in wave_update_llist_v1.2.txt. Please, also send me the list of lines that are not in v1.2 (independently of the naming convention); if these are few, then I will add them to v1.3 and we can update the naming convention if need be, if these are many, we may need to come up with a more efficient way to do this.

fcashman:
There are many, ~400. I can update the names fairly painlessly in the new file and send it to you. How do v1.2 and 1.3 interact with the ascii file, whose naming is not the same?

ntejos:
I believe the ascii file (latest version, e.g. v1.3) defines the name, and it uses the wavelength provided to look for atomic information from the parsing of the original catalogs (i.e. Morton, Verner, yours). @profxj may correct me if I'm wrong.

fcashman:
Okay, take a look at the file, I double checked it yesterday and hopefully there are no duplicates to v1.2. As of now, the names match the ascii file, but I can modify them over the weekend to keep the naming convention, and submit it to your pull request.

ntejos:
Sounds good for now. Another question regarding wave_update_llist_v1.2.txt. For OI 924 you provided a wavelength of 924.950 AA, while in the v1.2 of llist is is also listed at 924.95 AA...

fcashman:
Same line. Yes, the new Ritz wavelengths are accurate to 3 significant figures, so some are small changes. I wanted to be consistent and include where there were any changes in the 3rd decimal place.

PR #436, opened 2017-12-08
Add reference redshift option for stack plots

jnburchett:
Minor addition for plots.stack_plot to accept zref argument instead of using redshift for each AbsLine.  Useful for stack plots relative to a systemic redshifts.

### DIFF
--- a/linetools/analysis/plots.py
+++ b/linetools/analysis/plots.py
@@ -43,6 +43,7 @@ def stack_plot(abslines, vlim=[-300,300.]*u.km/u.s, nrow=6, show=True, spec=None
       If True, add EW values on the figure
     zref : float, optional
       Reference redshift for velocity scale; if None, each AbsLine.z is used
+
     Returns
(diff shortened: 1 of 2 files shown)


linetools/analysis/plots.py
```diff
@@ -81,7 +85,9 @@ def stack_plot(abslines, vlim=[-300,300.]*u.km/u.s, nrow=6, show=True, spec=None
         norm_sig = iline.analy['spec'].sig/co
         #
         # Plot
-        velo = iline.analy['spec'].relative_vel((1+iline.z)*iline.wrest)
+        if zref is None:
```
profxj: Include in the test in test_plots.py

jnburchett: Done.  Also, important point:  absline redshifts need to be set with AbsLine.setz() now since the vlims refactor.  A number of tests set redshifts in the old way.

PR #437, opened 2017-12-11
Adds info in error message for LSF interpolations

ntejos:
Added `wv0` information for error in LSF interpolation.


profxj:
merging.

PR #438, opened 2017-12-13
Creating line set llist_v1.3, adding SIV 1073 to it

ntejos:
Also, more info in an error in LSF added.


profxj:
@ntejos -- Please integrate with the line list provided by @fcashman in PR #434 
which I am going to merge now.

linetools/lists/sets/README
```diff
@@ -8,3 +8,5 @@ v1.1
 v1.2
   Adding some extra forbidden lines for Galaxy (NT Jun 28, 2017)
   Turned off distracting higher order Balmer lines in Galaxy
+v1.3
```
profxj: all good, but this needs to be synced with Cashman.

ntejos:
Roger, sounds good.

fcashman:
Hi, here are the renamed lines that were not in v1.2, which should go in v1.3.

[new_lines_llist_v1.3.txt](https://github.com/linetools/linetools/files/1568773/new_lines_llist_v1.3.txt)

ntejos:
Thanks @fcashman . Given the large amount of new lines (~400) I'm not sure what would be the best approach to add them to the new llist_v1.3. @profxj do you have a preferred approach?

profxj:
Nothing preferred @ntejos 

You can see what I've done previously.  I'd add all of them
to ISM but none of them to Strong unless you think otherwise.

profxj:
@ntejos -- What do we need to do to close this PR?
It is getting stale.

ntejos:
We need to solve conflicts on the current (in development) llist_v1.3.ascii. These conflicts must be fixed within `mk_sets.add_cashman17()`. Also, decide to which sets of lines individual cashman lines go (e.g. "Strong", "ISM", etc.)

profxj:
There are no lines with _ct in them in the current v1.3 linelist on this
branch.   Anyways, I am fussing with this now.

ntejos:
One sec, will push the commit then.

profxj:
too late

ntejos:
Ok

profxj:
Almost here.  A question or two for @fcashman --

I see a few lines where it seems your new wavelengths might
be lower precision than previous, e.g. NII 1083 where Morton
gives 1083.993700 and your list has 1083.990000.  Is your value
really better in this case?  (it is also nearly 0.004A different!)
Or less dramatic, this OI line:  950.884600 vs. 950.885000
Or something in between:  1402.772900 vs. 1402.770000

I want to get these 'right' if possible. 
Thanks

fcashman:
Hey, I am traveling today, leaving here (AAS) in 40 minutes for the airport. I'll take a good look at it when I'm back in the office. Promise!

profxj:
Hmm..  Hold up a bit. 

It is now my understanding that many of these revised
values are theoretical.  I've always been highly preferential
to lab values, even when the reported error is larger.

Am going to rethink this PR...

fcashman:
Okay, I dug a little deeper. So the theoretical N II lines on NIST come from Reader, Corliss, Wiese, & Martin 1980, with wavelengths with an uncertainty of 0.001 Ang. Both N II experimental and theoretical lines listed on NIST are identical. Morton uses N II experimental wavelengths verbatim from Eriksson 1983 with an uncertainty of 0.0020 Ang. For this specific wavelength, given the uncertainty, their value is close to overlapping despite the apparent difference, therefore you could certainly use the Eriksson value here. 

All O I wavelengths in Morton come from Moore's 1976 paper on O I. However, Moore released a paper for C, N, O atoms+ions in 1993. NIST superseded her 1976 paper with her 1993 update. As you pointed out, the values of 950.884600 vs. 950.885000 are almost the same, but we gave preference to new sources if it was from the same author. You could keep her older value and there'd likely negligible difference.

Both 3p level Si IV lines (1402 and 1393) wavelengths in Morton are experimental and come from Griesmann & Kling 2000, and interestingly it was a NIST experiment result. They also have values for C IV in this paper. These wavelengths look to be more accurate for both ions, and I wonder why NIST doesn't list them. It may be because they do stipulate that they agree with the existing data within the uncertainties. I could write to Gillian Nave and ask her. She was on the FTS project.

fcashman:
Regarding your later comment, I can make version of cashman17.ascii that only uses Ritz when there is not an observed wavelength. Let me know if that'd be more useful.

profxj:
@fcashman -- Let's have a vidcon to discuss further
when it is convenient for you (and me).  I propose next week.

For now, I am closing this PR and will address 2 or 3 small
items in a separate PR.

PR #440, opened 2017-12-14
AbsSystem and AbsComponent attributes

jnburchett:
Extends the AbsComponent object so that b values and velocity centroids are attribs.  Also, method to shift component velocities to systemic value.  Includes above attribs in component table.

### DIFF
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -142,7 +142,9 @@ class AbsComponent(object):
             slf.attrib['logN'] = logN
             slf.attrib['sig_logN'] = sig_logN
         else:
-            raise ValueError('The line measurements for the components are'
+            print(vels,cols,bs)
+            import pdb; pdb.set_trace()
+            raise ValueError('The line measurements for the components are '
                              'not consistent with one another.')
         # Return
         return slf
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -138,7 +138,12 @@ class AbsComponent(object):
             slf.attrib['sig_b'] = medberr * u.km/u.s
             slf.attrib['vel'] = medvel * u.km/u.s
             slf.attrib['velerr'] = medvelerr * u.km / u.s
-        logN,sig_logN  = ltaa.log_clm(slf)
+            logN,sig_logN  = ltaa.log_clm(slf)
+            slf.attrib['logN'] = logN
+            slf.attrib['sig_logN'] = sig_logN
+        else:
+            raise ValueError('The line measurements for the components are'
+                             'not consistent with one another.')
         # Return
         return slf
 
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -154,7 +154,7 @@ class AbsComponent(object):
             raise ValueError('The line measurements for the lines in this '
                              'component are not consistent with one another.')
             print(slf)
-        else:
+        elif medb != 0: # Empty b-values (upper lims) can throw off bcrit
             warnings.warn('The line measurements for the lines in this component'
                           ' are not consistent with one another.')
             print(slf)
--- a/linetools/isgm/abscomponent.py
+++ b/linetools/isgm/abscomponent.py
@@ -75,7 +75,7 @@ class AbsComponent(object):
     """
     @classmethod
     def from_abslines(cls, abslines, stars=None, comment="", reliability='none',
-                      tol=0.01, **kwargs):
+                      tol=0.01, chk_meas=False, **kwargs):
         """Instantiate from a list of AbsLine objects
 
         Parameters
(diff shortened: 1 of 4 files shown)


linetools/isgm/abscomponent.py
```diff
@@ -100,6 +118,48 @@ def from_abslines(cls, abslines, stars=None, comment="", reliability='none', **k
...
+        slf.attrib['logN'] = logN
+        slf.attrib['sig_logN'] = sig_logN
+        # Stop or throw warnings
+        if (bcrit&colcrit):
+            pass
+        elif chk_meas:
+            raise ValueError('The line measurements for the lines in this '
+                             'component are not consistent with one another.')
+            print(slf)
+        elif medb != 0: # Empty b-values (upper lims) can throw off bcrit
```
profxj: Are you sure medb should be zero?

linetools/isgm/abscomponent.py
```diff
@@ -100,6 +118,48 @@ def from_abslines(cls, abslines, stars=None, comment="", reliability='none', **k
...
+            warnings.warn('The line measurements for the lines in this component'
+                          ' are not consistent with one another.')
+            print(slf)
```
profxj: Maybe remove these print statements.

If you wish to keep them, add a verbose flag
and 

``` 
if verbose:
```

linetools/isgm/abscomponent.py
```diff
@@ -100,6 +118,48 @@ def from_abslines(cls, abslines, stars=None, comment="", reliability='none', **k
...
+        vels = np.array([al.attrib['vel'].value for al in abslines])
+        velerrs = np.array([al.attrib['sig_vel'].value for al in abslines])
+        medcol = np.median(cols)
+        medcolerr = np.median(colerrs)
+        medb = np.median(bs)
+        medberr = np.median(berrs)
+        medvel = np.median(vels)
+        medvelerr = np.median(velerrs)
+        # Perform checks on measurements
+        colcrit = all(np.abs(cols - medcol) / medcol < tol) == True
```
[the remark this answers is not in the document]
jnburchett: Fixed

linetools/isgm/abscomponent.py
jnburchett: Went the 'verbose' route.

linetools/isgm/abscomponent.py
jnburchett: Yeah, it's so that the code doesn't barf when upper limits go through the routine.

linetools/isgm/abscomponent.py
```diff
@@ -100,6 +118,48 @@ def from_abslines(cls, abslines, stars=None, comment="", reliability='none', **k
...
+        slf.attrib['sig_N'] = medcolerr / u.cm ** 2
+        slf.attrib['b'] = medb * u.km / u.s
+        slf.attrib['sig_b'] = medberr * u.km / u.s
+        slf.attrib['vel'] = medvel * u.km / u.s
+        slf.attrib['sig_vel'] = medvelerr * u.km / u.s
+        logN, sig_logN = ltaa.log_clm(slf.attrib)
+        slf.attrib['logN'] = logN
+        slf.attrib['sig_logN'] = sig_logN
+        # Stop or throw warnings
+        if (bcrit&colcrit):
```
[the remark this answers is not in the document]
jnburchett: I've cleaned this up a bit and clarified the intended operations.  Added else statement too.

jnburchett:
May we merge now, @profxj ?

ntejos:
I still don't fully understand this PR. I remember a time when I wanted to have Doppler parameters as attributes of AbsComponent objects, but @profxj preferred to have it inside the "attrib" dictionary of the AbsComponent instead. Why this later approach needs to be revisited? Also, the "velocity centroid" concept for an AbsComponent is a bit ambiguous to me... w/r to what redshift are these velocities defined? If it is the `z_comp` redshift then all these velocity centroids must be `=0` by definition and is pointless to have them.

profxj:
the b-values are still in attrib, but there is now a property way of accessing them.

profxj:
merging.

ntejos:
Wait... I'm fine with b-values as attributes. But what about the "velocity difference"? For it to make sense one would need to define a new redshift different than the redshift of the AbsComponent... that seems better placed in a object like AbsSystem rather than AbsComponent.

jnburchett:
@ntejos, I think I can explain from the perspective of using these objects in my workflow.  If I have a bunch of measurements in a line list, these may not have any 'systemic' redshift attached a priori.  Then I import the lines into the AbsLine, arrange these into AbsComponents, and only then would I start shuffling things into systems.  If I throw away the systemic redshifts given in the initial linelist, bookkeeping and sanity checking against the original linelist becomes VERY hairy.  There now exists a method to reset components within a system to all have velocities relative a system, so that's always an option.

PR #441, opened 2017-12-20
Attrib fixes

profxj:
Fussing with a few tests to address the last PR. 
Should have caught these before merging, but alas.


profxj:
Failing on the new numpy but I bet there will be fixes
to astropy that will deal with these.

ntejos:
Looks good to me.

PR #442, opened 2017-12-20
added script to get HST/COS LP from date + test

ntejos:
Script to get HST/COS LP from date. I found this convenient to have as a script.


profxj:
Looks fine.

Merge when tests pass on Travis.

PR #445, opened 2017-12-30
copy method for AbsSystem

jnburchett:
As stated, provides method for quick reinstantiation of AbsSystem objects.


linetools/isgm/abssystem.py
```diff
@@ -300,6 +300,33 @@ def add_component(self, abscomp, tol=0.2*u.arcsec,
...
+        newsys = self.__class__(self.coord, self.zabs, self.vlim, zem=self.zem,
+                           abs_type=self.abs_type, NHI=self.NHI,
+                           sig_NHI=self.sig_NHI, flag_NHI=self.flag_NHI,
+                           name=self.name)
+        newsys._components = [comp.copy() for comp in self._components]
+        newsys.kin = self.kin
+        newsys.ZH = self.ZH
+        newsys.sig_ZH = self.sig_ZH
+        newsys.flag_ZH = self.flag_ZH
+        newsys._EW = self._EW
```
profxj: I wasn't aware of this 'hidden' attribute _EW

Is it new?

linetools/isgm/abssystem.py
```diff
@@ -314,11 +341,12 @@ def chk_component(self, component):
         """
         return True
 
-    def fill_ionN(self, **kwargs):
+    def fill_ionN(self, vrange, **kwargs):
```
profxj: Is making vrange a new requirement going to break 
previous code?  It might..

profxj: Can you make it optional, i.e.

```
vrange=None
```

linetools/isgm/utils.py
```diff
@@ -419,7 +419,7 @@ def table_from_complist(complist):
     return tab
 
 
-def iontable_from_components(components, ztbl=None, NHI_obj=None):
+def iontable_from_components(components, ztbl=None, NHI_obj=None, vrange=None):
```
profxj: I would really prefer to deprecate this method that I wrote
in a really ugly way.  Is it possible to modify table_from_complist()
to do what you want?  I'm happy to help.

linetools/isgm/abssystem.py
ntejos: My feeling is that such attribute it should not exist at the `AbsSystem` level, it's an attribute of `AbsLine` instead

linetools/isgm/utils.py
jnburchett: Sure I think it's possible.  Could include a switch for ion table vs. component table and handle the two separately.  The ion table will not include b values or component velocities (and their errors) but will contain summed column densities.  Also, the modified table_from_complist() should contain someway to find the proper ion in the component list.

For the record, I'm not sure what is so ugly about iontable_from_components(), which leads me to believe that I might produce something no prettier.  Let's put our heads together on it?

profxj: Yes, let's discuss today.

jnburchett:
@profxj Given that AbsComponent is being refactored and to move things along with dependent PRs, may I suggest that we punt on deprecating ltiu.iontable_from_components()?  That leaves one loose end on this PR (abssystem.py, lines 317-320); please see my comment and advise.  Thanks!

profxj:
@jnburchett -- Let's resolve this conflict together.
Maybe tomorrow.  And let's get you to clone the main
repo instead of using your own fork.  That way, I can
do some of this work for you in the future.  Thanks.

jnburchett:
@profxj, I think this one's ready to fly from the nest.

profxj:
Passing all sensible tests.

Merging.

PR #453, opened 2018-01-16
Revise NiII f-value and add SVI

profxj:
As titled.

Uses Jenkins&Tripp 2006 for NiII 1317.

Adds SIV to set list.

Updates linelist.ascii file as appropriate.

No new tests or docs.


profxj:
Addresses Issue #453 
(which like this PR should read SIV not SVI).

PR #454, opened 2018-01-24
Llist qsos

ntejos:
Added LineList AGN with some of transition observed in AGN spectra. Docs and widget updated. 

Still some lines missing from https://github.com/profxj/xidl/blob/master/Spec/Lines/Lists/qso2.lst as there are not yet in the llist set. e.g. (OIV 1406, HeII 1640, NIII] 1750, NeIV 2423, NeV 3345, NeV 3426, FeVII 3587, HeII 4687).


profxj:
Looks great.  Merge if all but the
2 astropy tests fail in Travis.

PR #457, opened 2018-02-07
test for AbsComponent.copy()

ntejos:
Added test for copy() in AbsComponent object.


jnburchett:
Thanks...I should've done this in the first place!  Merging...

PR #458, opened 2018-02-09
Linetools IO for IGMGuesses

ntejos:
Some utilities for IO capabilities for IGMGuesses in pyigm


linetools/isgm/abscomponent.py
```diff
@@ -243,15 +243,21 @@ def from_dict(cls, idict, coord=None, **kwargs):
...
+        # slf = cls(radec, tuple(idict['Zion']), idict['zcomp'], Quantity(idict['vlim'], unit='km/s'),
 
         # backwards compatibility
-        for key in ['reliability', 'Reliability']:
-            if key in idict.keys():
-                reliability = idict[key]
-                break
-            else:
-                reliability = 'none'
+        if 'reliability' in idict.keys():
```
profxj: I can't see how the old code doesn't do exactly the same thing.
But you must have a reason..

linetools/isgm/io.py
```diff
@@ -179,3 +181,100 @@ def write_joebvp_from_components(comp_list, specfile, outfile):
         s = comp.repr_joebvp(specfile, flags=flags, b_default=b_val)  # still, b values from abslines take precedence if they exist
         f.write(s)
     f.close()
+
+
+def read_igmg_to_components(igmguesses_json):
```
profxj: IGMGuesses is a pyigm construct.
I think this method needs to go there.

And, I'll note that IGMSightline probably 
already does most of what you want here.

linetools/isgm/io.py
```diff
@@ -179,3 +181,100 @@ def write_joebvp_from_components(comp_list, specfile, outfile):
...
+        comp_list += [comp]
+    # sort list by redshift
+    redshifts = [comp.zcomp for comp in comp_list]
+    inds = np.argsort(redshifts)
+    comp_sorted = list(np.array(comp_list)[inds])
+    #return
+    return comp_sorted
+
+
+def write_igmg_from_components(comp_list, specfile, fwhm, outfile='IGM_model.json'):
```
profxj: Again, this should be in pyigm.  And 
make it work with IGMSightline which
I don't think has a write() method yet.

ntejos:
You are right, will close this PR and move the code to pyigm. Thanks.

linetools/isgm/abscomponent.py
ntejos: I found it clearer in the new code. but should be doing the same.

profxj:
merging.

PR #459, opened 2018-02-12
Two small, useful edits related to Components

profxj:
As titled.
Several IGMSurveys in pyigm were
crashing because of the coord speed-up.

No new tests or docs.


profxj:
Merging

PR #461, opened 2018-02-19
Compare Methods

profxj:
Preparing for a proper test of writing IGMGuesses files.

These methods will probably come in handy for other contexts.

New tests.  No new docs.


profxj:
@ntejos -- I am of the mind to strip out 'reliability' and 'comment' 
as default attributes of the AbsComponent class.  Am leaving them
for now, but consider yourself warned..

PR #465, opened 2018-03-01
Update astropy helpers

profxj:
removes an (old) bad line in setup.py

also updates ah_boostrap.py

Probably should have done so years ago  ;)


profxj:
merging.

PR #467, opened 2018-03-20
Fix oxygen

jnburchett:
The III->IV ionization energy was incorrect due to a duplicated entry in the ionenergy tuple for O.


ntejos:
Looks good to me.

profxj:
merging

PR #478, opened 2018-06-04
Abscomponent bug fix

jnburchett:
The from_dict_to_table() method had apparently been moved to linetools.lists.utils, but this had not been refactored in AbsComponent.add_abslines_from_linelist()


profxj:
If you can think of a real simple test to add,
please do.  That will insure we catch it next
time. But if not, no worries.  Merge away.

jnburchett:
Done!  Test added.

PR #479, opened 2018-07-05
Box_smooth option

profxj:
Adds an additional option for smoothing (that may
have used to exist).

### DIFF
--- a/linetools/spectra/xspectrum1d.py
+++ b/linetools/spectra/xspectrum1d.py
@@ -830,7 +830,7 @@ class XSpectrum1D(object):
         return velo
 
     #  Box car smooth
-    def box_smooth(self, nbox, preserve=True, **kwargs):
+    def box_smooth(self, nbox, preserve=True, scale_sig=True, **kwargs):
         """ Box car smooth the spectrum
 
         Parameters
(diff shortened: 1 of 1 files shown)


linetools/spectra/xspectrum1d.py
```diff
@@ -853,7 +855,9 @@ def box_smooth(self, nbox, preserve=True, **kwargs):
             from astropy.convolution import convolve, Box1DKernel
             new_fx = convolve(self.flux, Box1DKernel(nbox), **kwargs)
             if self.sig_is_set:
-                new_sig = convolve(self.sig, Box1DKernel(nbox), **kwargs) / np.sqrt(nbox)
+                new_sig = convolve(self.sig, Box1DKernel(nbox), **kwargs)
+                if scale_sig:
+                    new_sig /= np.sqrt(nbox)
```
jnburchett: Is decreasing the S/N like this commonplace? If used recklessly, this could lead to erroneously small S/N values and (further?) correlated errors within the spectrum.  At some point, I could see the virtue of a flag or something indicating that the spectrum has been smoothed (and thus the low S/N).   I suppose that the user should just really need to be more careful than this, though.

PR #482, opened 2018-08-21
Return Figure object for abssystem and emsystem

jnburchett:
As stated, trivial changes to allow Figure object passthrough for abssystem and emsystem when using return_fig option


profxj:
Merging

PR #485, opened 2018-11-06
limit sigfigs in z widget

mattcwilde:


profxj:
Not the most elegant fix, but ok.

PR #489, opened 2019-01-14
Integrate specdb and continuum fitting

profxj:
Integrates `linetools` with `specdb` but does not require the latter.

Minor updates to the continuum fitting code too.

Updated docs but no new tests.


profxj:
Approved and merging

PR #491, opened 2019-01-18
Log n attrib

ntejos:
modified AbsComponent.from_dict() and isgm.utils.table_from_complist() tosearch for logN info in the AbsComponent.attrib dictionary, and not the .logN attribute (deprecated).

+ extra docs in xspecgui.rst


linetools/isgm/utils.py
```diff
@@ -411,6 +411,12 @@ def table_from_complist(complist, summed_ion=False, NHI_obj=None, vrange=None,
     # Rename
     tab.rename_column('zcomp', 'z_comp')
 
+    # logN attribs, they should be stored only in the icomp.attrib dict()
+    for ak in  ['flag_N', 'logN', 'sig_logN']:
+        values = [icomp.attrib[ak] for icomp in complist]
+        import pdb; pdb.set_trace()
```
profxj: Need to remove this

profxj: I think you should add `flag_N`, `logN` etc to the list of 
attributes that occur a few lines down (e.g. including `b`, `sig_b`).
No need for a separate for loop here.

linetools/isgm/utils.py
```diff
@@ -402,7 +402,7 @@ def table_from_complist(complist, summed_ion=False, NHI_obj=None, vrange=None,
     tab['ion'] = [icomp.Zion[1] for icomp in complist]
 
     # . attributes (required ones)
-    for attrib in ['zcomp', 'Ej', 'flag_N', 'logN', 'sig_logN',
+    for attrib in ['zcomp', 'Ej',
```
profxj: Should likely remove `b`, `sig_b`, `vel` and `sig_vel` too

profxj:
@jnburchett -- Can you put eyes on this PR too?  
I worry it could adversely affect CASBaH if we are
doing something wrong here.

linetools/isgm/abscomponent.py
```diff
@@ -287,15 +287,16 @@ def from_dict(cls, idict, coord=None, skip_abslines=False, **kwargs):
                         slf.attrib[ak] = np.array(slf.attrib[ak])
 
         # Deprecated column (again)
-        if Ntup is not None:
+        if (Ntup is not None) and ('logN' not in idict['attrib'].keys()):
```
jnburchett: @profxj @profxj While testing, I was trying to instantiate a COSHalos object (pyigm.cgm.cos_halos), and the following line crashes:
`if (Ntup is not None) and ('logN' not in idict['attrib'].keys()):
`
It appears that the dict getting fed to the abscomponent.from_dict() function doesn’t have an ‘attrib’ key.  (`KeyError: 'attrib'`).  However, the dicts loaded from the json files for our CASBaH CGM survey do actually contain this key.  Thus, I'm starting to wonder if the issue is coming from deprecated formats for the COS-Halos survey json?

linetools/isgm/utils.py
ntejos: You are right. Modified accordingly.

linetools/isgm/abscomponent.py
ntejos: How about before creating the `'attrib'` key in the dictionary before feed it into `abscomponent.from_dict()`?

linetools/isgm/utils.py
ntejos: Agreed. The idea now is to impose all these attributes to go in the `.attrib` dict()

linetools/isgm/abscomponent.py
jnburchett: Not sure I totally understand when this would happen...do you mean before the survey data is output to json or as it's being loaded in?

ntejos: If the input dictionary (`idict`) to be fed to `abscomponent.from_dict()` does not have a key named `'attrib'` I suggest trying to create it before passing it to the method. For instance,

`idict['attrib'] = attrib_dict`

where `attrib_dict` is a dictionary with the attributes for AbsComponent (including LogN).

ntejos:
I see now. Will try to do what you suggest.
<http://www.fis.ucv.cl/ntejos>

ntejos:
@jnburchett could you check again if the error remains?

profxj:
@jnburchett -- If you approve we can merge this.

profxj:
merging

PR #492, opened 2019-01-24
Stackplot fixes and veeper/trident compatibility

jnburchett:
As stated.  Should make stackplot generation more customizable and keep up with other software packages that do or should lean heavily on linetools...not to name names here of course.


profxj:
merging

PR #493, opened 2019-02-03
Fix test_repr_joebvp Travis errors

jnburchett:
As stated, address some failed tests dealing with AbsComponent.


profxj:
Looks good.  Only failing the astropy development test.

PR #494, opened 2019-02-08
Remember wvlim in lt_xspec

ntejos:
As stated.

(Regarding remembering flux limits too, I think is better for xspec to guess the y-limits depending on each data than fixing it.)


profxj:
merging.

PR #495, opened 2019-02-08
Add continuum tag for spec I/O, update Install and travis

profxj:
As titled.

A few bits and pieces here.

Only test fail is astropy development.  That will
eventually get us but I'm ok with what we have for now.


profxj:
merging.

PR #499, opened 2019-02-28
Additional options for continuum fit

marijana777:
Adding options to:
- define input and output files with knots in lt_continuumfit 
- define continuum knots automatically, by including only pixels that do not correspond to absorption lines (code: analysis/continuumfnd.py )


linetools/analysis/continuumfnd.py
```diff
@@ -0,0 +1,206 @@
...
+        Identifies pixels that correspond to absorption lines, and do not includes them in continuum.
+
+        Parameters sm, npix, lchmin, ewsnlim that best define continuum might be different for different data sets.
+        Recomended using lt_continuumfit script (gui) after.
+
+    Parameters
+    ----------
+    spec : XSpectrum1D or str
+      input spectrum, or a file that contains input XSpectrum1D spectrum
+    sm : float
```
profxj: if the parameter is option, indicate it as such.

linetools/analysis/continuumfnd.py
```diff
@@ -0,0 +1,206 @@
...
+    flx3 = sp3.flux
+    sigf3 = sp3.sig
+
+
+    # 1) Define pixels that are on the continuum.
+    #    Smooth the spectrum, find pixels around maximums in the flux in the smoothed spectrum,
+    #    assume that the continuum is well defined by these pixels only.
+    #    This flux will be too high, and is used only in the step (1)
+    ipixs = []  # continuum pixels
+    for i in range(len(flx3)):
```
profxj: Why a nested set of `for` loops?  
This is not Pythonic.  Is there another way?

linetools/analysis/continuumfnd.py
```diff
@@ -0,0 +1,206 @@
...
+    #    Define EW = sum of (1 - flux/continuum) of these adjacent pixels
+    #    and a condition that EW > ewsnlim * sigma(EW) ; default ewsnlim = 2
+
+    # find pixels that correspond to absorption lines
+    ind0s = [] # pixels corresponding to absorption lines
+    imin = -1
+    iew = 0
+    isigew = 0
+
+    for i in range(len(flx)):
```
profxj: Gack, another `for` loop. 

Surely you can use logic to do this without a `for` loop.

linetools/analysis/continuumfnd.py
```diff
@@ -0,0 +1,206 @@
...
+            else:
+                if (iew > ewsnlim * isigew):
+                    for j in range(i - imin + 1):
+                        ind0s.append(j + imin)
+                iew = 0
+                isigew = 0
+                imin = -1
+
+    # Find pixels that do not correspond to absorption lines
+    ind0s = list(set(ind0s))
```
profxj: Why `list` manipulation.  Just use array + indices.

linetools/analysis/continuumfnd.py
marijana777: Indicated optional parameters

linetools/analysis/continuumfnd.py
marijana777: Now using one for loop and np.max. The code is also now faster

linetools/analysis/continuumfnd.py
marijana777: Now not using for loops. Instead, using np.cumsum, np.diff, and np.where. The code is also now faster.

linetools/analysis/continuumfnd.py
marijana777: Addressed with the previous change.

marijana777:
Addressed all requested changes for now.

profxj:
Thanks for this @marijana777 .  
The only thing left is to add a test or two of your `contknots` method.
Add it to the `tests` folder in `analysis`.

marijana777:
Thank you @profxj for the comments. Just added the test of the contknots method

profxj:
This all looks good now @marijana777 
But wait until PR #512 is merged into `master`
and then pull that in to yours so that Travis tests
should pass.  Once they do, I'll approve and merge.

profxj:
Ok, @marijana777 pull in `master` and confirm that Travis passes.
THanks.

marijana777:
@profxj confirming that Travis passed on test_continuumfnd.py in all cases. I also confirmed that the rest of the Travis output seems almost the same as in the most recent merged PR at this moment (i.e., the same tests failed, etc). 
Should we merge this PR now?

PR #508, opened 2019-08-20
Limits and stack_plot fix

jnburchett:
Correcting errors occurring when passing component lists to stack_plot and in extracting json from CGMSurvey.


profxj:
merging

PR #510, opened 2019-09-17
Tweak: Function to identify metals and interlopers in a given spectrum

ntejos:
As stated. A simple function that identifies pixels that belong to a metal line or a higher-order HI Lyman absorption. Useful for masking out interlopers.


ntejos:
Will move this to pyigm.surveys.utils.

PR #515, opened 2020-04-17
LSF G140L LP4

ntejos:
Implemented COS LSF G140L LP4. 

Only initialization tests have been added.


ntejos:
I have now added all the LP4 LSFs, including G130M and G160M...

PR #517, opened 2020-06-29
Add 'ANGSTROMS' unit to account for FUSE data in spectra/io.py.

jchowk:
FUSE NVO files in the MAST archive adopt the units of  'ANGSTROMS' (all upper case), which doesn't get captured in the present list of spectra/io.py. Adding this to line 305 fixes the issue.


profxj:
merging.

PR #520, opened 2020-09-21
Update dev version handling

finagle29:
This should make it so that, if a linetools sdist is made inside a linetools git repo, the correct version will go into version.py
Then, when installing from that sdist not in a linetools git repo, the correct version is pulled from version.py
This should make it so that the linetools conda-forge package will provide the correct value for linetools.__version__

I'll test the creation of and installation from sdists and report back results.


profxj (approved):
Thanks for this!

profxj:
Merging

PR #521, opened 2020-10-12
fussing with stack plot

mattcwilde:
matplotlib >3.0.3 produces error when trying to use 'step-mid' with the linestyle attribute as this was deprecated behavior.  

Also adding another way to make a stackplot from spec_file


profxj (approved):
thx!

profxj:
merge

PR #523, opened 2020-10-20
Actually fix dev version handling

finagle29:
Followup to #520 , this time with verification that the resulting sdists, when installed generate the correct `linetools.__version__` field


profxj (approved):
merging

PR #530, opened 2022-12-21
Avoid crash with overly long METADATA card

profxj:
An update to `astropy` appears to have it no longer
slurp in extra long CONTINUE's of strings.  This
lends to a crash when trying to read them in.

This PR offers a hack/fix as a temporary solution.


ntejos (approved):
works for me.

PR #531, opened 2023-02-14
Updates for Python v3.10 and refactoring of tests

profxj:
As titled.

We'll see how this goes...


finagle29 (approved):
all looks good to me

PR #538, opened 2024-01-12
imp -> importlib

ktchrn:
Replaced some more uses of imp with to importlib. Running tests with pytest doesn't flag anything up, though that was just with astropy 6.0.0 and numpy 1.26.3


profxj:
thx!

PR #540, opened 2024-09-12
Verbosity fix

jhennawi:
This minor PR suppresses unnecessary printing to the screen when performing basic operations.


profxj (approved):
thx!

PR #541, opened 2024-10-29
Pkg resources

SunilSimha:
As titled, `pkg_resources` has been deprecated for some time and finally broken in Python 3.12. This PR also rids the `frb` repository of some errors related to the `linetools` dependency.


profxj (approved):
thx for this!

PR #543, opened 2025-06-20
Radec utility string that points to a visibility plot 

ntejos:
Radec utility string that points to a visibility plot -- it adds to the convenience strings already available. The default site is LCO and the default day is today, so the user must update these two manually on the website itself. A warning on this is added.


profxj (approved):
thanks for this!

